PR #2, opened 2019-02-21
ENH: Vacuum widgets

hhslepicka:
## Description

This PR adds the vacuum widgets following the specifications provided.
It also add one little smoke test and initial documentation.
More tests to come in the near future.

## Motivation and Context

This addresses the need for widgets for vacuum screens.

## How Has This Been Tested?

Locally tested.

Docstrings available and it will be deployed to gh-pages as soon as merged into master.


hhslepicka:
Changed to allow_failures with Python 3.7 and pcds-tag while Python 3.7 packages are not available.
See https://github.com/pcdshub/pcds-recipes/issues/27

pcdswidgets/symbols/gauges.py
```diff
@@ -0,0 +1,502 @@
...
+        HotCathodeGauge [interlocked="false"] #interlock {
+            border: 0px;
+        }
+        HotCathodeGauge [state="Error"] #icon {
+            qproperty-penColor: red;
+            qproperty-penWidth: 2;
+        }
+
+    """
+    Q_ENUMS(ContentLocation)
```
teddyrendahl: Why is this not defined on the base class?

pcdswidgets/symbols/base.py
```diff
@@ -0,0 +1,316 @@
...
+        self.layout().addWidget(self.interlock)
+
+    def clear(self):
+        """
+        Remove all inner widgets from the interlock frame layout.
+        """
+        layout = self.interlock.layout()
+        if not layout:
+            return
+        while layout.count() != 0:
```
teddyrendahl: End up copying this a lot of places. Might be nice to have a PyDM utility

pcdswidgets/symbols/base.py
```diff
@@ -0,0 +1,316 @@
...
+        self.icon = None
+
+        self._show_icon = True
+        self._show_status_tooltip = True
+        self._icon_size = -1
+        self._controls_location = ContentLocation.Bottom
+        self.setup_ui()
+
+    @Property(str)
+    def channelsPrefix(self):
```
teddyrendahl: A little confused by this. You aren't actually connecting to a PV? Why do we need the `ca` prefix?

pcdswidgets/symbols/gauges.py
```diff
@@ -0,0 +1,502 @@
...
+            color: gray;
+        }
+        PiraniGauge [state="On"] {
+            qproperty-brush: green;
+            color: black;
+        }
+
+    """
+    Q_ENUMS(ContentLocation)
+    NAME = "Pirani Gauge"
```
teddyrendahl: Where does this `NAME` go?

pcdswidgets/symbols/pumps.py
```diff
@@ -0,0 +1,545 @@
...
+    def create_channels(self):
+        """
+        Method invoked when the channels associated with the widget must be
+        created.
+        This method also sets the channel address for the control button.
+        """
+        super(TurboPump, self).create_channels()
+        if self._channels_prefix:
+            self.start_stop_btn.channel = "{}{}".format(self._channels_prefix,
+                                                        ":RUN_SW")
```
teddyrendahl: Thinking about pulling all these random suffixes out into class attributes. That we don't have to go hunting.

i.e
```python
class TurboPump:
    _command_suffix = ':RUN_SW'
```

pcdswidgets/symbols/pumps.py
```diff
@@ -0,0 +1,545 @@
...
+        super(TurboPump, self).assemble_layout()
+        if self.controlsLocation in [ContentLocation.Top,
+                                     ContentLocation.Bottom]:
+            self.start_stop_btn.orientation = Qt.Horizontal
+            self.start_stop_btn.setMinimumSize(100, 40)
+        else:
+            self.start_stop_btn.orientation = Qt.Vertical
+            self.start_stop_btn.setMinimumSize(100, 80)
+
+    def create_channels(self):
```
teddyrendahl: Also wondering if you could do more of the `tooltip` automatically from here. I'm a little worried about making sure that every channel updates the status correctly, and doing it on `create_channels` makes some sense to me?

pcdswidgets/symbols/gauges.py
teddyrendahl: The tooltip

pcdswidgets/symbols/gauges.py
hhslepicka: By a Qt limitation in which if you have it at the base class it works at the Designer but when the UI file is imported it complains about not being able to reach the ENUM and the only way that we could get it to work was to have it at the widget class.

pcdswidgets/symbols/base.py
hhslepicka: Sure! We can add that there as a utility function to clean a layout.

pcdswidgets/symbols/base.py
hhslepicka: We need the protocol as whatever you configure at the `channelsPrefix` is concatenated with the suffix for the mixins.
Here is the usage:
- At the widget, configure the channelsPrefix to something like: `ca://VALVE`
- The interlock channel will be set to: `ca://VALVE:OPN_OK`

pcdswidgets/symbols/pumps.py
hhslepicka: I'm not sure that I understood this one completely.

pcdswidgets/symbols/pumps.py
hhslepicka: Addressed at the latest commit.

pcdswidgets/symbols/gauges.py
```diff
@@ -0,0 +1,517 @@
...
+        self.controls_layout = QVBoxLayout()
+        self.controls_layout.setSpacing(0)
+        self.controls_layout.setContentsMargins(0, 0, 0, 0)
+        self.controls_frame.setLayout(self.controls_layout)
+        self.controls_frame.layout().addWidget(self.pressure_label)
+
+        self.assemble_layout()
+        self.update_status_tooltip()
+
+    @Property(ContentLocation)
```
teddyrendahl: Still not sure why this property is repeated on all the widgets?

pcdswidgets/symbols/base.py
```diff
@@ -0,0 +1,342 @@
...
+        Returns
+        -------
+        str
+        """
+        if hasattr(self, 'NAME'):
+            return self.NAME
+        else:
+            return ""
+
+    def destroy_channels(self):
```
teddyrendahl: Could we generalize this by keeping track of all the channels in an attribute like the `PyDMWidgetBase` does? Then this looks like:

```python
def destroy_channels(self):
    for channel in self._channels:
        ...
```

pcdswidgets/symbols/pumps.py
```diff
@@ -0,0 +1,630 @@
...
+
+    def destroy_channels(self):
+        """
+        Method invoked when the channels associated with the widget must be
+        destroyed.
+        This method also clears the channel address for the pressure label and
+        control button.
+        """
+        super(IonPump, self).destroy_channels()
+        self.pressure_label.channel = None
```
teddyrendahl: Sometimes we `disconnect` sometimes we set the address to `None`, concerned why we need both and why in separate places?

examples/show_icon.py
```diff
@@ -0,0 +1,26 @@
+import os
+from qtpy.QtWidgets import QApplication, QWidget
+
+import pcdswidgets.icons as icons
+
+def screenshot(widget, filename):
```
teddyrendahl: `mily` 😸

pcdswidgets/symbols/gauges.py
teddyrendahl: Seeing this again, fine copying the `Q_ENUMS` I suppose but do we need to copy the `Property` everywhere as well?

pcdswidgets/symbols/gauges.py
hhslepicka: Because by some weird reason the Qt Designer does not like the property not being with the Q_ENUMS definition. I tried it and it just collapses.
I would recommend to just move ahead with this property being duplicated for now until we find a better way to handle it.

pcdswidgets/symbols/base.py
hhslepicka: I believe we can. I will change it.

pcdswidgets/symbols/pumps.py
hhslepicka: We disconnect when we create the channel. We send `None` when it is a widget that is used. The PyDM widget in this case will search for the channel and do the proper disconnection. 
I think that it is cleaner than hijacking the channels property and calling disconnect.

examples/show_icon.py
hhslepicka: It can be moved there in the future... it is a simple ✂️ , 📋

pcdswidgets/symbols/gauges.py
hhslepicka: This problem was solved by creating the plugin for the base class so Qt can register the hierarchy properly.

pcdswidgets/symbols/pumps.py
```diff
@@ -0,0 +1,522 @@
...
+        self.controls_layout = QVBoxLayout()
+        self.controls_layout.setSpacing(0)
+        self.controls_layout.setContentsMargins(0, 0, 0, 0)
+        self.controls_frame.setLayout(self.controls_layout)
+        self.controls_frame.layout().addWidget(self.start_stop_btn)
+
+        self.assemble_layout()
+        self.update_status_tooltip()
+
+    def assemble_layout(self):
```
teddyrendahl: I see this repeated quite a few times. Maybe another mixin or something?

hhslepicka: Addressed with the latest commit.

PR #10, opened 2019-04-29
BUG: Use widget size and not event for paint event

domitto:
## Description
When the widget is used inside a QScrollArea the paint event pass in a smaller or bigger size than the widget resulting in a bad icon.
This fix ensure that the widget size is used when painting to ensure proper behaviour.


teddyrendahl:
Thanks for the PR! I want @hhslepicka to have a look, but seems to make a lot of sense to me.

PR #18, opened 2019-08-08
FIX: QtDesigner started to present issues with properties returning None

hhslepicka:
## Description

Today after updating my code and using Qt newer than 5.6 I started having problems to load the widgets. I narrowed it down to the fact that properties for the widgets were returning None.
I believe that Qt no longer likes it.

## How Has This Been Tested?

Locally since we can't test QtDesigner at CI.

N/A

### DIFF
--- a/pcdswidgets/vacuum/others.py
+++ b/pcdswidgets/vacuum/others.py
@@ -53,12 +53,12 @@ class RGA(PCDSSymbolBase):
 
     @Property(str, designable=False)
     def channelsPrefix(self):
-        pass
+        return super(RGA, self).channelsPrefix
 
     @Property(bool, designable=False)
     def showIcon(self):
-        pass
+        return super(RGA, self).showIcon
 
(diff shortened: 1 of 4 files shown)


pcdswidgets/vacuum/others.py
```diff
@@ -53,12 +53,12 @@ def sizeHint(self):
 
     @Property(str, designable=False)
     def channelsPrefix(self):
-        pass
+        return super(RGA, self).channelsPrefix
```
klauer: Still 2.7-style `super()` calls?

hhslepicka: Ooops... Sorry about that. I will push a fix.

hhslepicka: Fixed with the latest commit https://github.com/pcdshub/pcdswidgets/pull/18/commits/5fcc4221bb6c144abde513f2123b1006ce0fc441

PR #39, opened 2020-03-17
ENH: Add new property ophydClass at BaseWidget

hhslepicka:
## Description
This PR adds a new property called `ophydClass` at `PCDSSymbolBase` that will allow users to opt-out of the default suggested Ophyd Device class to be used for the expert screen.

## Motivation and Context
This is a change requested by @n-wbrown .

## How Has This Been Tested?
Tests are passing.

## Screenshots (if appropriate):
Screen Shot 2020-03-17 at 4 26 27 PM

### DIFF
--- a/pcdswidgets/vacuum/base.py
+++ b/pcdswidgets/vacuum/base.py
@@ -50,6 +50,8 @@ class PCDSSymbolBase(QWidget, PyDMPrimitiveWidget, ContentLocation):
         self._icon_size = -1
         self._icon = None
 
+        self._ophyd_class = ""
+
         self.interlock = QFrame(self)
         self.interlock.setObjectName("interlock")
         self.interlock.setSizePolicy(QSizePolicy.Expanding,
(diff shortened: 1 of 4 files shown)


pcdswidgets/vacuum/base.py
```diff
@@ -270,6 +272,34 @@ def rotateIcon(self, rotate):
         if self.icon:
             self.icon.rotation = angle
 
+    @Property(str)
+    def ophydClass(self):
```
klauer:
```suggestion
    def expertOphydClass(self):
```
Maybe? It's OK as-is.

pcdswidgets/vacuum/gauges.py
```diff
@@ -75,6 +74,7 @@ def __init__(self, parent=None, **kwargs):
             readback_suffix=self._readback_suffix,
             readback_name='pressure',
             **kwargs)
+        self.ophydClass = "pcdsdevices.gauge.GaugePLC"
```
klauer: Consider keeping the `OPHYD_CLASS` in the class body.

Then you can generically do a:
```suggestion
        self.ophydClass = self.OPHYD_CLASS
```

pcdswidgets/vacuum/base.py
hhslepicka: I was aiming for a small name... It for sure looks better with the `expert` at the name...

hhslepicka: I will adopt this name as well as modify the class body variable to `EXPERT_OPHYD_CLASS` so it becomes a 1-1 mapping

pcdswidgets/vacuum/gauges.py
hhslepicka: Addressed with https://github.com/pcdshub/pcdswidgets/pull/39/commits/44e502889dad60eafb881740eaec571897461802

pcdswidgets/vacuum/base.py
hhslepicka: Addressed with https://github.com/pcdshub/pcdswidgets/pull/39/commits/44e502889dad60eafb881740eaec571897461802

hhslepicka:
New screenshot:
Screen Shot 2020-03-17 at 5 11 40 PM

PR #45, opened 2020-06-24
FIX: Bug at ophydExpertClass setter and update class for FastShutter

hhslepicka:
## Description
- Fix a bug when trying to set the class name for a widget
- Update the FastShutter expert screen class as requested by @slacAdpai 

## Motivation and Context
Fix a bug and update the class to the newly created one.

### DIFF
--- a/pcdswidgets/vacuum/base.py
+++ b/pcdswidgets/vacuum/base.py
@@ -413,7 +413,7 @@ class PCDSSymbolBase(QWidget, PyDMPrimitiveWidget, ContentLocation):
         prefix = remove_protocol(self.channelsPrefix)
         klass = self.expertOphydClass
         if not klass:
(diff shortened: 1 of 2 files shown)


pcdswidgets/vacuum/base.py
```diff
@@ -413,7 +413,7 @@ def _handle_icon_click(self):
         prefix = remove_protocol(self.channelsPrefix)
         klass = self.expertOphydClass
         if not klass:
-            logger.error('No ophydClass specified for pcdswidgets %s',
+            logger.error('No experOphydClass specified for pcdswidgets %s',
```
klauer, line 416:
```suggestion
            logger.error('No expertOphydClass specified for pcdswidgets %s',
```

PR #67, opened 2021-07-23
ENH: add sortable/filterable table inspired by pmps-ui

ZLLentz:
## Description

Add the `FilterSortWidgetTable` widget. This widget allows us to repeat a template generically in a table form, and handle things like sorting and filtering in a re-usable way.

Here's a screenshot of the TMO screen I am working on that uses this widget:

image

Features:

- [x] Allow a repeated widget from a template and macros
- [x] Dynamically allow sorting on any value
- [x] Allow the user to register filters and enable/disable them live
- [x] Allow the user to manually rearrange the rows of the table (check the configure box)
- [ ] Allow the user to hide/show rows individually
- [ ] Allow the user to save/load the table arrangement
- [ ] Support resizing to increase the size of the internal widgets (can we do this easily?)
- [ ] Unit tests

## Motivation and Context

This is a common motif and we should be able to re-use it in multiple contexts

## How Has This Been Tested?

Only interactiveley so far

Only docstrings so far

### DIFF
--- a/pcdswidgets/table.py
+++ b/pcdswidgets/table.py
@@ -427,7 +427,7 @@ class FilterSortWidgetTable(QtWidgets.QTableWidget):
         else:
             self.showRow(row)
 
-    def activate_filter(self, active: bool, filter_name: str) -> None:
+    def activate_filter(self, filter_name: str, active: bool) -> None:
         """
         Activate or deactivate a filter by name.
 
--- a/pcdswidgets/table.py
+++ b/pcdswidgets/table.py
@@ -36,6 +36,7 @@ class FilterSortWidgetTable(QTableWidget):
         self._filters = {}
         self._initial_sort_header = 'index'
         self._initial_sort_ascend = True
+        self._hide_headers = []
 
         # Table settings
         self.setShowGrid(True)
@@ -224,6 +225,8 @@ class FilterSortWidgetTable(QTableWidget):
         sort_menu = menu.addMenu('Sorting')
         for header_name in self._header_map.keys():
             if header_name == 'widget':
+                continue
+            if header_name in self.hide_headers_in_menu:
                 continue
             inner_menu = sort_menu.addMenu(header_name.lower())
             asc = inner_menu.addAction('Ascending')
@@ -424,6 +427,17 @@ class FilterSortWidgetTable(QTableWidget):
         """
         self.sort_table(self.initial_sort_header, self.initial_sort_ascending)
 
+    @QtCore.Property('QStringList')
+    def hide_headers_in_menu(self):
+        """
+        A list of headers that we don't want to see in the sort menu.
+        """
+        return self._hide_headers
+
+    @hide_headers_in_menu.setter
+    def hide_headers_in_menu(self, headers):
+        self._hide_headers = headers
+
     def sort_table(self, header, ascending):
         """
         Rearrange the ordering of the table based on any of the value fields.
@@ -468,7 +482,7 @@ class FilterSortWidgetTable(QTableWidget):
         for row in range(self.rowCount()):
             header.moveSection(header.visualIndex(row), row)
 
-    @QtCore.Property(bool)
+    @QtCore.Property(bool, designable=False)
     def configurable(self):
         """
         Whether or not the table can be manipulated from the UI.
@@ -476,7 +490,7 @@ class FilterSortWidgetTable(QTableWidget):
         If True, the table rows can be dragged/dropped/rearranged.
         If False, the table rows can no longer be selected.
 
-        This begins as False if unset.
+        This begins as False if unset and can be changed in the context menu.
         """
         return self._configurable
 
--- a/pcdswidgets/table.py
+++ b/pcdswidgets/table.py
@@ -1,8 +1,10 @@
+from __future__ import annotations
+
 import dataclasses
 import functools
 import logging
 import json
-import typing
+from typing import Any, Callable, Optional
 
 from pydm.utilities import is_qt_designer
 from pydm.widgets import PyDMEmbeddedDisplay
@@ -10,6 +12,8 @@ from pydm.widgets.channel import PyDMChannel
 from qtpy import QtCore, QtGui, QtWidgets
 
 logger = logging.getLogger(__name__)
+
+FilterFunc = Callable[[dict[str, Any]], bool]
 
 
 class FilterSortWidgetTable(QtWidgets.QTableWidget):
@@ -19,6 +23,25 @@ class FilterSortWidgetTable(QtWidgets.QTableWidget):
     This will allow you to sort or filter based on macros and based on the
     values in each pydm widget.
     """
+    # Public instance variables
+    template_widget: PyDMEmbeddedDisplay
+
+    # Private instance variables
+    _ui_filename: Optional[str]
+    _macros_filename: Optional[str]
+    _macros: list[dict[str, str]]
+    _channel_headers: list[str]
+    _macro_headers: list[str]
+    _header_map: dict[str, int]
+    _channels: list[PyDMChannel]
+    _filters: dict[str, FilterInfo]
+    _initial_sort_header: str
+    _initial_sort_ascend: bool
+    _hide_headers: list[str]
+    _configurable: bool
+    _watching_cells: bool
+
+
     def __init__(self, *args,  **kwargs):
         super().__init__(*args, **kwargs)
         self._ui_filename = None
@@ -47,14 +70,14 @@ class FilterSortWidgetTable(QtWidgets.QTableWidget):
 
         self._watching_cells = False
 
-    def channels(self):
+    def channels(self) -> list[PyDMChannel]:
         """
         Tell PyDM about our table channels so it knows to close them at exit.
         """
         return self._channels
 
     @QtCore.Property(str)
-    def ui_filename(self):
+    def ui_filename(self) -> str:
         """
         Name of the ui file that is to be repeated to fill the table.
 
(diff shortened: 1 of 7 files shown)


ZLLentz:
This is collecting dust. I'm going to make the tests pass, ask for a review, merge, and move the rest of the todo items into an issue.

ZLLentz:
A big uncommitted comment block I found on my NFS clone, I must have been trying to figure out the remaining bullet points:
```
+# For showing/hiding individual rows (filter exceptions), we can
+# Add an extra column at the start that unhides when we enter config mode
+# The extra column has a combobox with the show/hide/auto options
+# When we filter, we check this combobox first. Show is always shown, etc.
+# For saving, we can probably just need:
+# - macros for each row in the current logical order
+# - the current logical-visual mapping
+# - the current show/hide/auto
+# - the current filter options
+# For loading, we create the rows from the macros file in order,
+# then move the visual ordering of the header to match the previous mapping.
+# Saving and loading can be in the right-click menu
+# For increasing the size on larger monitors... I have no idea.
```

ZLLentz:
Seems to still work using the TMO screen, though the performance isn't great when you have a lot of items and/or ask for new sorts.

ZLLentz:
Created #70 with follow-up information

ZLLentz:
This closes #69

pcdswidgets/table.py
```diff
@@ -0,0 +1,596 @@
...
+                    # If inactive, record it as unfiltered/shown
+                    show_row.append(True)
+            if all(show_row):
+                self.showRow(row)
+            else:
+                self.hideRow(row)
+        else:
+            self.showRow(row)
+
+    def activate_filter(self, active, filter_name):
```
klauer, line 368: API: other methods have `filter_name` first, it seems

pcdswidgets/table.py
```diff
@@ -0,0 +1,596 @@
...
+        function.
+        """
+        return self._initial_sort_header
+
+    @initial_sort_header.setter
+    def initial_sort_header(self, header):
+        self._initial_sort_header = header
+        if not is_qt_designer():
+            # Do a sort after a short timer
+            # HACK: this is because it doesn't work if done immediately
```
klauer, line 407: Because the other widgets have yet to be initialized, I assume?

pcdswidgets/table.py
```diff
@@ -0,0 +1,596 @@
...
+        """
+        self.reset_manual_sort()
+        if ascending:
+            order = QtCore.Qt.AscendingOrder
+        else:
+            order = QtCore.Qt.DescendingOrder
+        col = self._header_map[header]
+        self.sortItems(col, order)
+        if not self.isSortingEnabled():
+            # This triggers a sort
```
klauer, line 462: Looking at the qt docs, I wonder if `sortByColumn(col, order)` would have sufficed?

pcdswidgets/table.py
```diff
@@ -0,0 +1,596 @@
...
+            # Set up the data columns and the channels
+            for header in self._channel_headers:
+                source = widget.findChild(QtCore.QObject, header)
+                item = ChannelTableWidgetItem(
+                    header=header,
+                    channel=source.channel,
+                    )
+                self.setItem(row_position, index, item)
+                self._header_map[header] = index
+                self._channels.append(item.pydm_channel)
```
klauer, line 202: `pydm_channel` is technically optional for `ChannelTableWidgetItem`; does `None` matter here? (I think it does, based on `disconnect()` calls later...)

klauer (approved):
I didn't consider what might be contained in your follow-up PR, but only what I saw in this PR.

All suggestions are minor and just code smell / potential bug-related. 

Regardless of that, the functionality the widget provides is really nice. I expect we'll be able to use it for atef and other projects.

ZLLentz:
Oh man, sorry this is so messy. I'll have to come back to this one before merging I guess... it's really awkward because it's deployed as-is in dev.

klauer:
I don't think any of what I suggested are really so bad, but rather have places for easy improvements. I'd be totally OK with merging/tagging and letting you circle back to this if/when you get time.

ZLLentz:
I kind of disagree, I don't like to compromise on code quality. The important thing to cut off here to ensure it gets merged is the scope, not the small tweaks.

ZLLentz:
Ok, this PR next. To-do:
- Add an example that I can run offline that uses the table widget. If I don't have enough time for unit tests I need to at least make time to run the code again later in a controlled way, and maybe the example will be useful on its own.
- Address all of Ken's review comments

pcdswidgets/table.py
```diff
@@ -0,0 +1,596 @@
...
+        self.reinit_table()
+
+    def reload_ui_file(self):
+        """
+        Load the UI file and inspect it for PyDM channels.
+        """
+        try:
+            self.template_widget.filename = self.ui_filename
+        except Exception:
+            logger.exception('')
```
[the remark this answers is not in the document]
ZLLentz: Yeah this definitely needs to return and have "something" as the error message

pcdswidgets/table.py
```diff
@@ -0,0 +1,596 @@
...
+        self._header_map = {}
+        if not self._macros and self._channel_headers:
+            return
+        # Column 1 displays widget, 2 is index, the rest hold values
+        ncols = 2 + len(self._channel_headers) + len(self._macro_headers)
+        self.setColumnCount(ncols)
+        for col in range(1, ncols):
+            self.hideColumn(col)
+        for macros in self._macros:
+            widget = PyDMEmbeddedDisplay(parent=self)
```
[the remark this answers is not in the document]
ZLLentz: ah yep that's a pretty huge block with sub-indents

pcdswidgets/table.py
```diff
@@ -0,0 +1,596 @@
...
+from pydm.utilities import is_qt_designer
+from pydm.widgets import PyDMEmbeddedDisplay
+from pydm.widgets.channel import PyDMChannel
+from qtpy import QtCore, QtGui, QtWidgets
+
+logger = logging.getLogger(__name__)
+
+
+class FilterSortWidgetTable(QtWidgets.QTableWidget):
+    """
```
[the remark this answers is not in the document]
ZLLentz: I've now realized that I imported typing here, but then hardly used it

pcdswidgets/table.py
```diff
@@ -0,0 +1,596 @@
...
+
+        Parameters
+        ----------
+        macros_list : list of dict
+            A list where each element is a dictionary that defines the macros
+            to pass in to one instance of the repeated widget. All dicts must
+            have the same keys or this will not work properly.
+        """
+        self._macros = macros_list
+        self._macro_headers = list(self._macros[0].keys())
```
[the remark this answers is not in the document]
ZLLentz: Confirmed that the table should still run with empty macros list (resulting in a very boring table)

pcdswidgets/table.py
ZLLentz: yeah this definitely only works as-is because of the previously misplaced disconnect that only runs once, on the first item's channel

pcdswidgets/table.py
```diff
@@ -0,0 +1,596 @@
...
+        configure_action = menu.addAction('Configure')
+        configure_action.setCheckable(True)
+        configure_action.setChecked(self.configurable)
+        configure_action.toggled.connect(self.request_configurable)
+        active_sort_action = menu.addAction('Active Re-sort')
+        active_sort_action.setCheckable(True)
+        active_sort_action.setChecked(self.isSortingEnabled())
+        active_sort_action.toggled.connect(self.setSortingEnabled)
+        sort_menu = menu.addMenu('Sorting')
+        for header_name in self._header_map.keys():
```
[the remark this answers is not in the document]
ZLLentz: Is it preferable to use the implied keys here? It's not obvious to me which is better, one is more readable and the other is more succinct.

pcdswidgets/table.py
```diff
@@ -0,0 +1,596 @@
...
+        ----------
+        row : int
+            The row index to inspect. 0 is the current top row.
+        """
+        if self._filters:
+            values = self.get_row_values(row)
+            show_row = []
+            for filt_info in self._filters.values():
+                if filt_info.active:
+                    show_row.append(filt_info.filter_func(values))
```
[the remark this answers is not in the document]
ZLLentz: You're absolutely right, the default should be "True" to show. I ran into this problem while setting up the example and making lots of errors.

pcdswidgets/table.py
ZLLentz: I guess so, unfortunately I don't actually remember the consequence and I haven't documented it here...

pcdswidgets/table.py
ZLLentz: Yeah, you're absolutely correct. It is sufficient to call only that method here instead of this whole block and the previous method, along with using `setSortingEnabled` separately to enable/disable the live re-sorting for changing values.

Not sure how I fell into this trap... must have either missed this or misinterpreted the behavior at some point.

pcdswidgets/table.py
ZLLentz: I think is is caused by some combination of the designer attributes being applied in some undefined order and the values used for the sorting not being immediately available.

For now I'm going to expand the comment...

pcdswidgets/table.py
```diff
@@ -0,0 +1,596 @@
...
+                self.setItem(row_position, index, item)
+                self._header_map[header] = index
+                self._channels.append(item.pydm_channel)
+                index += 1
+
+        self._watching_cells = True
+        self.cellChanged.connect(self.handle_item_changed)
+        self.update_all_filters()
+
+    def add_context_menu_to_children(self, widget):
```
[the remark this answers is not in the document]
ZLLentz: I kind of agree, but considering this is only called once and only has two lines I'm more inclined to tear out the function entirely than to change the API of it

pcdswidgets/table.py
ZLLentz: Thanks, it turns out that switching this has no consequences because it's always called using kwargs only

ZLLentz:
I think I've gotten all the biggest fixes done. I'm going to merge this and move on.

PR #73, opened 2022-06-08
REF/ENH: designer widgets via entrypoints

klauer:
## Description
* Use the new entrypoint support in PyDM to export our widgets to the qt designer
* Marked as WIP as the categories are not yet set for the widgets and the original `plugin.py` remains

## Motivation and Context
* My local environment isn't as good as our deployed pcds-env and I don't want to have to tweak PYQTDESIGNERPATH any more

## How Has This Been Tested?
`pip install -e` and `Designer`


klauer:
(Restarting CI)

klauer:
I think this should be good now:
image

tangkong (approved):
LGTM.  As Ken knows I didn't get this working locally, but that's a me problem and shouldn't hold up the merge

ZLLentz:
The implementation looks correct
This is always a pain to double-check locally, so instead I'll trust Ken's test and compare against the widgets available before this change as a quick completeness sanity check:

image

Compared to previous, the new implementation:
- fixes the non-alphabetical order
- moves the gauges into the symbols category

I think I'll take a moment to find the bit that miscategorizes the gauges but the rest looks good

pcdswidgets/vacuum/gauges.py
```diff
@@ -1,12 +1,11 @@
-from qtpy.QtCore import QSize
```
ZLLentz, line 1: This file missing _qt_designer_ attributes to preserve the "PCDS Gauges" category

klauer: Good catch 👍

ZLLentz (approved):

PR #77, opened 2022-11-30
Allow for custom ui file to be open when widget icon is pressed.

C0bit:
## Description

When device widget icon is pressed a predefined screen will appear as a tab in the typhos screen. If no screen is defined typhos works as usual.

## Motivation and Context

Allows for more information to be displayed about a device, such as interlock information.
https://jira.slac.stanford.edu/browse/LCLSECSD-1735

## How Has This Been Tested?

Made a test screen that contains a pcds widget and checked the behavior when the icon is pressed. 
Tested variation of inputs in designer properties.

### DIFF
--- a/pcdswidgets/vacuum/base.py
+++ b/pcdswidgets/vacuum/base.py
@@ -7,8 +7,7 @@ from pydm.widgets.channel import PyDMChannel
 from qtpy.QtCore import Q_ENUMS, Property, QSize
 from qtpy.QtGui import QCursor, QPainter
 from qtpy.QtWidgets import (QFrame, QHBoxLayout, QSizePolicy, QStyle,
-                            QStyleOption, QVBoxLayout, QWidget, QStackedWidget,
-                            QComboBox, QTabWidget)
+                            QStyleOption, QVBoxLayout, QWidget, QTabWidget)
 
 from ..utils import refresh_style
 
--- a/pcdswidgets/vacuum/base.py
+++ b/pcdswidgets/vacuum/base.py
@@ -7,7 +7,7 @@ from pydm.widgets.channel import PyDMChannel
 from qtpy.QtCore import Q_ENUMS, Property, QSize
 from qtpy.QtGui import QCursor, QPainter
 from qtpy.QtWidgets import (QFrame, QHBoxLayout, QSizePolicy, QStyle,
-                            QStyleOption, QVBoxLayout, QWidget, QTabWidget)
+                            QStyleOption, QVBoxLayout, QWidget)
 
 from ..utils import refresh_style
 
@@ -443,6 +443,7 @@ class PCDSSymbolBase(QWidget, PyDMPrimitiveWidget, ContentLocation):
         try:
             import typhos
             from pydm.widgets import PyDMEmbeddedDisplay
+            from qtpy.QtWidgets import QTabWidget
 
         except ImportError:
             logger.error('Typhos not installed. Cannot create display.')
--- a/pcdswidgets/vacuum/base.py
+++ b/pcdswidgets/vacuum/base.py
@@ -4,11 +4,13 @@ import os
 from pydm.utilities import IconFont, remove_protocol
 from pydm.widgets.base import PyDMPrimitiveWidget
 from pydm.widgets.channel import PyDMChannel
+from pydm.widgets.embedded_display import PyDMEmbeddedDisplay
 from qtpy.QtCore import Q_ENUMS, Property, QSize
 from qtpy.QtGui import QCursor, QPainter
 from qtpy.QtWidgets import (QFrame, QHBoxLayout, QSizePolicy, QStyle,
-                            QStyleOption, QVBoxLayout, QWidget)
+                            QStyleOption, QVBoxLayout, QWidget, QTabWidget)
 
+from itertools import zip_longest
 from ..utils import refresh_style
 
 logger = logging.getLogger(__name__)
@@ -82,12 +84,12 @@ class PCDSSymbolBase(QWidget, PyDMPrimitiveWidget, ContentLocation):
         self.update_status_tooltip()
 
 
-        self.ui_file_path = []
+        self.ui_file_paths = []
         self.ui_file_macros = []
         self.ui_file_titles = []
 
-        self.tabWidget = None
-        self.embeddedDisplay = None
+        self.tab_widget = None
+        self.embedded_displays = list()
 
 
     def sizeHint(self):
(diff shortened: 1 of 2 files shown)


pcdswidgets/vacuum/base.py
```diff
@@ -432,6 +442,9 @@ def _handle_icon_click(self):
 
         try:
             import typhos
+            from pydm.widgets import PyDMEmbeddedDisplay
+            from qtpy.QtWidgets import QTabWidget
```
ZLLentz, line 446: `pydm` and `qtpy` are listed as requirements of `pcdswidgets`, so we don't need to hide them away inside of this try/except block which mostly functions to allow `pcdswidgets` to run without `typhos` (see the error message in the except clause). These imports belong at the top of the file instead.

pcdswidgets/vacuum/base.py
```diff
@@ -416,7 +424,9 @@ def _handle_icon_click(self):
                          self.__class__.__name__)
             return
 
-        if self._expert_display is not None:
+        if self.tabWidget is not None:
+            self.tabWidget.show()
```
ZLLentz, line 428: I think this part of the code that checks if we already have an open screen needs to behave more like it did previously.

The pre-existing non-tabbed branch of this if statement calls `raise_()` here to bring the open display to the front, in case it is already visible but is hidden behind another window. It also calls `return` to avoid going through the rest of the function and perhaps accidentally creating another widget copy.

pcdswidgets/vacuum/base.py
```diff
@@ -441,8 +454,62 @@ def _handle_icon_click(self):
         self._expert_display = display
         display.destroyed.connect(self._cleanup_expert_display)
 
-        if display:
-            display.show()
+        if len(self.ui_file_path) > 0:
+            self.tabWidget = QTabWidget()
+            self.embeddedDisplay = list()
+            self.tabWidget.setTabPosition(2)
```
ZLLentz, line 460: When working with enums it's usually best practice to reference the enum name instead of the value to make the code more readable.
```suggestion
            self.tabWidget.setTabPosition(QTabWidget.TabPosition.West)
```

pcdswidgets/vacuum/base.py
```diff
@@ -82,6 +81,15 @@ def __init__(self, parent=None, **kwargs):
         self.assemble_layout()
         self.update_status_tooltip()
 
+
+        self.ui_file_path = []
+        self.ui_file_macros = []
+        self.ui_file_titles = []
+
+        self.tabWidget = None
+        self.embeddedDisplay = None
```
ZLLentz, line 90: These names don't follow pep8 https://peps.python.org/pep-0008/#function-and-variable-names

pcdswidgets/vacuum/base.py
```diff
@@ -441,8 +454,62 @@ def _handle_icon_click(self):
         self._expert_display = display
         display.destroyed.connect(self._cleanup_expert_display)
 
-        if display:
-            display.show()
+        if len(self.ui_file_path) > 0:
+            self.tabWidget = QTabWidget()
+            self.embeddedDisplay = list()
```
ZLLentz, line 459: This naming breaks my brain: I assumed `embeddedDisplay` would be a single instance of `PyDMEmbeddedDisplay`

pcdswidgets/vacuum/base.py
```diff
@@ -441,8 +454,62 @@ def _handle_icon_click(self):
...
+                    title = files
+                else:
+                    title = self.ui_file_titles[i]
+
+                if i >= len(self.ui_file_macros):
+                    macros = ''
+                else:
+                    macros = self.ui_file_macros[i]
+
+                embedded.set_macros_and_filename(files, macros)
```
ZLLentz, line 476: Is files a variable that represents multiple files?

pcdswidgets/vacuum/base.py
```diff
@@ -441,8 +454,62 @@ def _handle_icon_click(self):
...
+
+                if i >= len(self.ui_file_titles):
+                    title = files
+                else:
+                    title = self.ui_file_titles[i]
+
+                if i >= len(self.ui_file_macros):
+                    macros = ''
+                else:
+                    macros = self.ui_file_macros[i]
```
ZLLentz, line 474: There's an, optional, more compact way to express this that you should know about using `itertools.zip_longest`:
```suggestion
            for files, title, macros in itertools.zip_longest(
                self.ui_file_path,
                self.ui_file_title,
                self.ui_file_macros,
            ):
                embedded = PyDMEmbeddedDisplay()
                title = title or files
                macros = macros or ''
```

pcdswidgets/vacuum/base.py
```diff
@@ -82,6 +81,15 @@ def __init__(self, parent=None, **kwargs):
         self.assemble_layout()
         self.update_status_tooltip()
 
+
+        self.ui_file_path = []
```
ZLLentz, line 85: Is this another variable with confusing plurality?

pcdswidgets/vacuum/base.py
```diff
@@ -441,8 +454,62 @@ def _handle_icon_click(self):
...
+    def ui_macros(self):
+        return self.ui_file_macros
+
+    @ui_macros.setter
+    def ui_macros(self, macros):
+        if macros != self.ui_macros:
+            self.ui_file_macros = macros
+
+    @Property('QStringList')
+    def titles(self):
```
ZLLentz, line 505: Should this last one maybe be `ui_titles` to match the others? I think seeing `titles` in designer would be confusing.

pcdswidgets/vacuum/base.py
```diff
@@ -441,8 +454,62 @@ def _handle_icon_click(self):
...
+                self.embeddedDisplay.append(embedded)
+
+
+            self.tabWidget.show()
+
+        elif display:
+            display.show() 
+    
+    @Property('QStringList')
+    def ui_path(self):
```
ZLLentz, line 487: And should this also be plural for clarity?

ZLLentz (commented):
The design choice of using tabs for this is great and the logic looks good here. I have one functional suggestion below, and then all of my other comments are questions about code clarity to make sure the code is easy to read in the future.

The automated tests are failing on pre-commit only (flake8, etc.)

ZLLentz (approved):
The code looks good.
I gave this a test run to finish out the review. My notes:

- Simple to set this up in designer
- Works as expected with a real device
- Was confused a bit because the typhos button doesn't do anything when the channel is empty. I'm not sure if this is the correct behavior but it's not related to this PR, it was always how this button worked.

I'm marking this as approved and clicking merge!

ZLLentz:
Hold up- actually, I'm not clicking merge because the automated test failed. Let's take a look.

ZLLentz:
It failed on basic internet connectivity breaking during the test runner- I'll re-run the failing test elements.

ZLLentz:
Something is wrong with our cloud CI tests- I ran the equivalent test locally and it passed, so I'll merge this despite the apparent cloud failure.

PR #82, opened 2023-02-27
ENH: add eps widget for EPS PLC structure

C0bit:
## Description

Widget for displaying EPS structure information as a ByteIndicator. The widget automatically sets the correct size and labels based on EPS structure PLC code, given the base PV of the EPS structure PV.

## Motivation and Context

[Jira Ticket](https://jira.slac.stanford.edu/browse/LCLSPC-373)

## How Has This Been Tested?

Using a PLC program with the modified twincat-general and twincat-motion, a test program was written. The test program was ran on the tst-motion PLC and its associated IOC. Using these PVs the eps-widget was tested to see if the widget functioned as expected.

image

### DIFF
--- a/pcdswidgets/eps_byteindicator.py
+++ b/pcdswidgets/eps_byteindicator.py
@@ -1,0 +1,155 @@
+from pydm.widgets import PyDMByteIndicator, PyDMChannel
+from qtpy.QtCore import Property, Qt
+from qtpy.QtGui import QColor
+from qtpy.QtWidgets import QTabWidget, QVBoxLayout, QWidget
+
+
+class ByteIndicator_NegativeNums(PyDMByteIndicator):
+    """
+        Modified Byte Indicator calss. Overides update_indicators
+            fucntion to allow for negative numbers to work for ByteIndicator
+    """
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def update_indicators(self):
+        if self._shift < 0:
+            value = int(self.value) << abs(self._shift)
+        else:
+            value = int(self.value) >> self._shift
+
+        bits = [(value >> i) & 1
+                for i in range(self._num_bits)]
+        for bit, indicator in zip(bits, self._indicators):
+            if self._connected:
+                if self._alarm_state == 3:
+                    c = self._invalid_color
+                else:
+                    c = self._on_color if bit else self._off_color
+            else:
+                c = self._disconnected_color
+            indicator.setColor(c)
+
+
+class EPSByteIndicator(QWidget):
+    """
+        Widget for displaying EPS interlocks
+    """
+
+    _qt_designer_ = {
+        "group": "PCDS Utilities",
+        "is_container": False,
+    }
+
+    template_widget = ByteIndicator_NegativeNums
+
+    _channels_prefix = None
+    _value_channel = None
+    _label_channel = None
+
+    _value_pv = ""
+    _label_pv = ""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.template_widget = ByteIndicator_NegativeNums()
+        layout = QVBoxLayout()
+        layout.addWidget(self.template_widget)
+        self.setLayout(layout)
+
+    def value_change(self, new_value):
+        """
+            Callback function when value changes
+        """
+        self.template_widget.value = new_value
+        self.template_widget.update_indicators()
+
+    def label_change(self, new_labels):
+        """
+            Callback function when the lables change
+        """
+        labels = "".join(map(chr, new_labels))
+        labels = labels.partition('\x00')
+        labels = labels[0].split(';')
+
+        self.template_widget.numBits = len(labels)
+        self.template_widget.labels = labels
+        self.template_widget.update_indicators()
+
+    @Property(str)
+    def channel(self):
+        """
+            PV of Base EPS strcture
+        """
+        return self._channels_prefix
+
+    @channel.setter
+    def channel(self, ch):
+        """
+            Set PV of Base EPS strcture
+        """
+        if ch != self._channels_prefix:
+            self._value_pv = ch + ":EPS:nFlags_RBV"
+            self._label_pv = ch + ":EPS:sFlagDesc_RBV"
+
+            _value_channel = PyDMChannel(address=self._value_pv,
+                                         value_slot=self.value_change)
+
+            _label_channel = PyDMChannel(address=self._label_pv,
+                                         value_slot=self.label_change)
+            try:
+                _value_channel.connect()
+                _label_channel.connect()
+            except Exception:
+                print('could not connect to PV')
+                self.template_widget._connected = False
+
+            self._channels_prefix = ch
+            self.template_widget._connected = True
+
+    @Property(bool)
+    def circles(self):
+        return self.template_widget.circles
+
+    @circles.setter
+    def circles(self, b):
+        self.template_widget.circles = b
+
+    @Property(QTabWidget.TabPosition)
+    def label_position(self):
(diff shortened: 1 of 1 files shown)


pcdswidgets/eps_byteindicator.py
```diff
@@ -0,0 +1,155 @@
...
+
+class ByteIndicator_NegativeNums(PyDMByteIndicator):
+    """
+        Modified Byte Indicator calss. Overides update_indicators
+            fucntion to allow for negative numbers to work for ByteIndicator
+    """
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def update_indicators(self):
```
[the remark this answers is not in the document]
ZLLentz: I also want to compare with https://github.com/pcdshub/pmps-ui/blob/74a8bbfe03ed90eb04b0b1370d3e761f8ebff6ab/widgets.py#L377-L397 from one of my open PRs

pcdswidgets/eps_byteindicator.py
```diff
@@ -0,0 +1,155 @@
...
+        """
+        self.template_widget.value = new_value
+        self.template_widget.update_indicators()
+
+    def label_change(self, new_labels):
+        """
+            Callback function when the lables change
+        """
+        labels = "".join(map(chr, new_labels))
+        labels = labels.partition('\x00')
```
ZLLentz, line 72: I'm surprised you had to map the ascii waveform and snip the null terminator yourself, that seems like a deficiency in pydm

ZLLentz: It looks like internally PyDM uses this helper function when it needs to do any sort of conversion like this: https://github.com/slaclab/pydm/blob/573e5b0c8dcfbd8e7bc8f7b8ba450b0b729c38e0/pydm/widgets/display_format.py#L27

pcdswidgets/eps_byteindicator.py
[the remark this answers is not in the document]
ZLLentz: I _think_ the only thing changed here compared to the normal `PyDMByteIndicatorWidget` is removing the `value<0` check (I don't understand why this is there in PyDM).

Does this work as intended? I suspect the negative numbers don't display properly but I haven't tried this specific patch.

The reason we get negative numbers here is that EPICS doesn't have unsigned types and has a maximum integer size of 32 bits, so when we send through an integer like 16#FFFFFFFF it gets interpreted as -1 instead of 2**32-1

pcdswidgets/eps_byteindicator.py
```diff
@@ -0,0 +1,155 @@
+from pydm.widgets import PyDMByteIndicator, PyDMChannel
+from qtpy.QtCore import Property, Qt
+from qtpy.QtGui import QColor
+from qtpy.QtWidgets import QTabWidget, QVBoxLayout, QWidget
+
+
+class ByteIndicator_NegativeNums(PyDMByteIndicator):
+    """
+        Modified Byte Indicator calss. Overides update_indicators
+            fucntion to allow for negative numbers to work for ByteIndicator
```
ZLLentz, line 10: Typos and indentation
Python docstrings should be intended to match the triple quote
```suggestion
    Modified Byte Indicator class.
    
    Overrides update_indicators function to allow for negative numbers to work for PyDMByteIndicator.
```

pcdswidgets/eps_byteindicator.py
```diff
@@ -0,0 +1,155 @@
...
+    """
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
ZLLentz, line 13: There's no need to override the init here if nothing has changed
```suggestion
```

pcdswidgets/eps_byteindicator.py
```diff
@@ -0,0 +1,155 @@
...
+            _value_channel = PyDMChannel(address=self._value_pv,
+                                         value_slot=self.value_change)
+
+            _label_channel = PyDMChannel(address=self._label_pv,
+                                         value_slot=self.label_change)
+            try:
+                _value_channel.connect()
+                _label_channel.connect()
+            except Exception:
+                print('could not connect to PV')
```
ZLLentz, line 104: Counterintuitively, this won't throw any sort of exception for an invalid or unreachable PV, so the error message here isn't accurate. Any exception raised here would be an internal PyDM error.

This actually starts a background thread that connects to the PV and forwards values/metadata to the main thread via qt signals. So if the channel can't connect, you just get `connection=False` metadata until it does, rather than an exception raised here.

If you need the connection state of the PV you can include a `connection_slot` in the `PyDMChannel` (expects a `bool`, `True` when the PV connects and `False` when the PV disconnects)

pcdswidgets/eps_byteindicator.py
```diff
@@ -0,0 +1,155 @@
...
+            self._channels_prefix = ch
+            self.template_widget._connected = True
+
+    @Property(bool)
+    def circles(self):
+        return self.template_widget.circles
+
+    @circles.setter
+    def circles(self, b):
+        self.template_widget.circles = b
```
ZLLentz, line 116: It's not a big deal when the function is two lines, but in general you should avoid single-letter variable names for readability.

pcdswidgets/eps_byteindicator.py
```diff
@@ -0,0 +1,155 @@
...
+                else:
+                    c = self._on_color if bit else self._off_color
+            else:
+                c = self._disconnected_color
+            indicator.setColor(c)
+
+
+class EPSByteIndicator(QWidget):
+    """
+        Widget for displaying EPS interlocks
```
ZLLentz, line 36: When you have a chance, you should go through and dedent all of your docstrings by 4 spaces
```suggestion
    Widget for displaying EPS interlocks
```

pcdswidgets/eps_byteindicator.py
```diff
@@ -0,0 +1,155 @@
...
+    """
+
+    _qt_designer_ = {
+        "group": "PCDS Utilities",
+        "is_container": False,
+    }
+
+    template_widget = ByteIndicator_NegativeNums
```
ZLLentz, line 44: This seems like a typo, did you mean to set a type annotation?
```suggestion
    template_widget: ByteIndicator_NegativeNums
```

ZLLentz (commented):
I found a bunch of small things, but overall I think this is really good.

pcdswidgets/eps_byteindicator.py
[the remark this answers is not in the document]
C0bit: In my testing, negative numbers displayed properly using this change.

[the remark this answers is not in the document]
ZLLentz: I read the PyDM source more closely and I see why this works now, so it's perplexing that the negative case is blocked in the original source. I'll investigate that separately later.

pcdswidgets/eps_byteindicator.py
```diff
@@ -0,0 +1,156 @@
...
+        return self._channels_prefix
+
+    @channel.setter
+    def channel(self, ch):
+        """
+        Set PV of Base EPS strcture
+        """
+        if ch != self._channels_prefix:
+            self._value_pv = ch + ":EPS:nFlags_RBV"
+            self._label_pv = ch + ":EPS:sFlagDesc_RBV"
```
ZLLentz, line 98: Does this need to be modified slightly after https://github.com/pcdshub/lcls-twincat-general/pull/67/commits/9b10a875c45bd579abb83e61697b6ce42344c97c ?

ZLLentz (approved):
I think this should be good to go
I'm going to go through and do the merging/tagging for the PLC-side pull requests and then merge this one afterwards

PR #92, opened 2025-03-10
Added text box to base vac widget

TrumanPar:
## Description
In short, I added a text box (QLabel) to the base vac widget along with several signals/sources to configure this box in designer (text size, text position, hide text). 

## Motivation and Context
This change is motivated from ECSENG-731. Maggie and Jing do not prefer the current method of adding a separate text box for each device on the vac screens.

## How Has This Been Tested?
This has been tested in my dev environment on existing vacuum screens. I have also demo'd this to Jing and she gave me the green light. Since the text box is hidden by default, I can confirm this does not graphically impact any existing widgets. I can also confirm that the control buttons are still drawn to the screen correctly. 

Currently I haven't described my changes elsewhere but please tell me where I should.

### DIFF
--- a/pcdswidgets/vacuum/base.py
+++ b/pcdswidgets/vacuum/base.py
@@ -6,9 +6,9 @@ from pydm.utilities import IconFont, remove_protocol
 from pydm.widgets.base import PyDMPrimitiveWidget
 from pydm.widgets.channel import PyDMChannel
 from pydm.widgets.embedded_display import PyDMEmbeddedDisplay
-from qtpy.QtCore import Q_ENUMS, Property, QSize
+from qtpy.QtCore import Q_ENUMS, Property, QSize, Qt
 from qtpy.QtGui import QCursor, QPainter
-from qtpy.QtWidgets import (QFrame, QHBoxLayout, QSizePolicy, QStyle,
+from qtpy.QtWidgets import (QLabel, QFrame, QHBoxLayout, QSizePolicy, QStyle,
                             QStyleOption, QTabWidget, QVBoxLayout, QWidget)
 
 from ..utils import refresh_style
@@ -60,6 +60,10 @@ class PCDSSymbolBase(QWidget, PyDMPrimitiveWidget, ContentLocation):
         self._icon_size = -1
         self._icon = None
 
+        self.name = None
+        self._show_name = False
+        self._font_size = 16
+
         self._icon_cursor = self.setCursor(
             QCursor(IconFont().icon("file").pixmap(16, 16))
         )
@@ -70,6 +74,14 @@ class PCDSSymbolBase(QWidget, PyDMPrimitiveWidget, ContentLocation):
         self.interlock.setObjectName("interlock")
         self.interlock.setSizePolicy(QSizePolicy.Expanding,
                                      QSizePolicy.Expanding)
+	
+        self.name = QLabel(self)
+        self.name.setWordWrap(True)
+        self.name.setSizePolicy(QSizePolicy.Maximum, 
+                                QSizePolicy.Maximum)
+        self.name.setVisible(self._show_name)
+        self.name.setAlignment(Qt.AlignCenter)
+        self.name.setStyleSheet(f"font-size: {self._font_size}px; background: transparent")
 
         self.controls_frame = QFrame(self)
         self.controls_frame.setObjectName("controls")
@@ -79,8 +91,12 @@ class PCDSSymbolBase(QWidget, PyDMPrimitiveWidget, ContentLocation):
         self.layout().setSpacing(0)
         self.layout().setContentsMargins(0, 0, 0, 0)
         self.layout().addWidget(self.interlock)
+
         if not hasattr(self, '_controls_location'):
             self._controls_location = ContentLocation.Bottom
+        if not hasattr(self, '_text_lcoation'):
+            self._text_location = ContentLocation.Top
+
         self.setup_icon()
         self.assemble_layout()
         self.update_status_tooltip()
@@ -126,6 +142,30 @@ class PCDSSymbolBase(QWidget, PyDMPrimitiveWidget, ContentLocation):
             self._controls_location = location
             self.assemble_layout()
 
+    @Property(ContentLocation)
+    def textLocation(self):
+        """
+        Property controlling where the PV name is displayed relative to the icon
+
+        Returns
+        -------
+        location : ContentLocation
+        """
+        return self._text_location
+    
+    @textLocation.setter
+    def textLocation(self, location):
+        """
+        Property controlling where the PV name is displayed relative to the icon
+
+        Parameters
+        ----------
+        location : ContentLocation
+        """
+        if location != self._text_location:
+            self._text_location = location
+            self.assemble_layout()
+
     @Property(str)
     def channelsPrefix(self):
         """
@@ -151,12 +191,14 @@ class PCDSSymbolBase(QWidget, PyDMPrimitiveWidget, ContentLocation):
         Parameters
         ----------
         prefix : str
-            The prefix to be used for the channels.
+           The prefix to be used for the channels.
         """
         if prefix != self._channels_prefix:
             self._channels_prefix = prefix
             self.destroy_channels()
             self.create_channels()
+            format_name = self._channels_prefix.split("//")[-1].split(":")[-3:]
+            self.name.setText(f"{format_name[0]}-\n{'-'.join(format_name[1:])}") # (":".join(self._channels_prefix.split("//")[-1].split(":")[-3:]))
 
     @property
     def icon(self):
@@ -197,6 +239,56 @@ class PCDSSymbolBase(QWidget, PyDMPrimitiveWidget, ContentLocation):
                 self.icon.setVisible(self._show_icon)
             self.assemble_layout()
 
+    @Property(bool)
+    def showName(self):
+        """
+        Whether or not to show the name when rendering the widget.
+
+        Returns
+        -------
+        bool
+        """
+        return self._show_name
+
+    @showName.setter
+    def showName(self, value):
+        """
+        Whether or not to show the name when rendering the widget.
+
+        Returns
+        -------
+        bool (set to true will display the name)
+        """
+        if value != self._show_name:
+            self._show_name = value
+            if self.name:
+                self.name.setVisible(self._show_name)
+            self.assemble_layout()
+
+    @Property(int)
+    def fontSize(self):
+        """
+        Set the font size for the name when rendering the widget.
+
+        Returns
+        -------
+        int
+        """
+        return self._font_size
+
+    @fontSize.setter
+    def fontSize(self, value):
+        """
+        Set the font size for the name when rendering the widget.
+
+        Returns
+        -------
+        int
+        """
+        if value != self._font_size:
+            self._font_size = value
+            self.name.setStyleSheet(f"font-size: {self._font_size}px; background: transparent")
+  
     @Property(bool)
     def showStatusTooltip(self):
         """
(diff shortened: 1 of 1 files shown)


pcdswidgets/vacuum/base.py
```diff
@@ -70,6 +74,14 @@ def __init__(self, parent=None, **kwargs):
         self.interlock.setObjectName("interlock")
         self.interlock.setSizePolicy(QSizePolicy.Expanding,
                                      QSizePolicy.Expanding)
+	
+        self.name = QLabel(self)
+        self.name.setWordWrap(True)
+        self.name.setSizePolicy(QSizePolicy.Maximum, 
+                                QSizePolicy.Maximum)
+        self.name.setVisible(self._show_name)
```
tangkong, line 82: Is it possible to only create this label if self._show_name is True?  As is, we might just be creating widgets that take up memory while hidden.

pcdswidgets/vacuum/base.py
```diff
@@ -151,12 +191,14 @@ def channelsPrefix(self, prefix):
...
         prefix : str
-            The prefix to be used for the channels.
+           The prefix to be used for the channels.
         """
         if prefix != self._channels_prefix:
             self._channels_prefix = prefix
             self.destroy_channels()
             self.create_channels()
+            format_name = self._channels_prefix.split("//")[-1].split(":")[-3:]
+            self.name.setText(f"{format_name[0]}-\n{'-'.join(format_name[1:])}") # (":".join(self._channels_prefix.split("//")[-1].split(":")[-3:]))
```
tangkong, line 201: Why are we manually adding newlines here?  Is the label's word-wrap no sufficient?

tangkong (commented):
Hi!  Thanks for the contribution!  It appears you have broken some of the tests in the process of making this change, we'll have to modify those tests so that they check the new functionality and pass.

Aside from that a few notes

pcdswidgets/vacuum/base.py
TrumanPar: The label's word wrap will only wrap the last ":" delimited field, which does not reflect how we currently format the manual text boxes. From what I've seen, we typically have a <br> to always wrap 2 ":" delimited fields on a newline. Likewise, my goal here is to also force 2 ":" fields on the newline, however I don't feel strongly one way or another.

pcdswidgets/vacuum/base.py
TrumanPar: I can initialize self.name to type None, and in the showIcon setter, I can do a check if it is None and initialize it to a QLabel if so. What do you think?

TrumanPar: Also thanks for the review Robert 😄

pcdswidgets/vacuum/base.py
```diff
@@ -60,6 +60,10 @@ def __init__(self, parent=None, **kwargs):
         self._icon_size = -1
         self._icon = None
 
+        self.name = None
```
ZLLentz, line 63: Nitpick: it seems like this `self.name=None` call is always unused in the current implementation. I think you should either make use of this as you suggested elsewhere or remove it.

pcdswidgets/vacuum/base.py
```diff
@@ -151,12 +191,14 @@ def channelsPrefix(self, prefix):
         Parameters
         ----------
         prefix : str
-            The prefix to be used for the channels.
+           The prefix to be used for the channels.
```
ZLLentz, line 194: Nitpick: possible accidental space deletion here? We probably should revert this.
```suggestion
            The prefix to be used for the channels.
```

pcdswidgets/vacuum/base.py
```diff
@@ -70,6 +74,14 @@ def __init__(self, parent=None, **kwargs):
...
         self.interlock.setSizePolicy(QSizePolicy.Expanding,
                                      QSizePolicy.Expanding)
+	
+        self.name = QLabel(self)
+        self.name.setWordWrap(True)
+        self.name.setSizePolicy(QSizePolicy.Maximum, 
+                                QSizePolicy.Maximum)
+        self.name.setVisible(self._show_name)
+        self.name.setAlignment(Qt.AlignCenter)
+        self.name.setStyleSheet(f"font-size: {self._font_size}px; background: transparent")
```
ZLLentz, line 84: Comment: I think using pixel font size is a good call, it will ensure consistent scaling with the widget pixel size :+1:

pcdswidgets/vacuum/base.py
ZLLentz: Nitpick: you should remove the trailing comment if you're done experimenting with different variants here.

Questions/food for thought:
- Is this text exactly what we want to include in the label?
- Does it work for all possible channel inputs without breaking?
- Is there ever a case where a user might want different text in the label with a manual override?
- Does it match what is done all or many existing labels? (specifically, the manual newline vs the auto line wrap)

I'm asking all these questions because split/join chains are a bit annoying to maintain, so if a simpler `prefix.replace(":", "-")` would work it is preferable. If not, then I at least want some guarantees that this still works with bad/strange inputs (such as those without `//` or pvs with very few `:`)

I'm also trying to head off the case where someone wants to display "something else" and ends up making a separate label anyway.

pcdswidgets/vacuum/base.py
```diff
@@ -79,8 +91,12 @@ def __init__(self, parent=None, **kwargs):
         self.layout().setSpacing(0)
         self.layout().setContentsMargins(0, 0, 0, 0)
         self.layout().addWidget(self.interlock)
+
         if not hasattr(self, '_controls_location'):
             self._controls_location = ContentLocation.Bottom
+        if not hasattr(self, '_text_lcoation'):
+            self._text_location = ContentLocation.Top
```
ZLLentz, line 98: Nitpick: It looks like you're repeating an existing pattern, but whoever did this before you seems to have been confused. I'm not sure how these could have ever been set prior to this line.

pcdswidgets/vacuum/base.py
ZLLentz: One last tiny tiny thing I noticed: this implementation doesn't match the gif you posted in the jira, which doesn't have any dashes. Not sure if this is meaningful in terms of testing done or meaningless.

pcdswidgets/vacuum/base.py
```diff
@@ -368,24 +460,87 @@ def assemble_layout(self):
...
+                ContentLocation.Bottom: (QVBoxLayout,
+                                        [grouped_frame,
+                                        self.controls_frame]),
+                ContentLocation.Left: (QHBoxLayout,
+                                    [self.controls_frame,
+                                        grouped_frame]),
+                ContentLocation.Right: (QHBoxLayout,
+                                        [grouped_frame,
+                                        self.controls_frame]),
+            }
```
ZLLentz, line 538: Food for thought: is there a way to implement this without creating similar variants of this dict in three places? Using a mapping was elegant when there was only one, but now that there are three it looks like a maintenance hazard. The logic is correct (I think?) but I wince a little when I think about modifying it.

pcdswidgets/vacuum/base.py
ZLLentz, line 1: One last request: is it possible to add a test for this new feature to https://github.com/pcdshub/pcdswidgets/blob/master/pcdswidgets/tests/vacuum/test_base.py?

ZLLentz (commented):
I think this is good, despite my nitpicks.

Main takeaways should be:

- Is it possible to simplify the logic for future maintainability?
- Is it possible to write a unit test that verifies this works as you expect?

Also, note that this breaks one of the existing tests. Is the existing test no longer valid (i.e. it was written with assumptions that are no longer true), or has some of the base functionality been broken?

pcdswidgets/vacuum/base.py
TrumanPar: I'll remove the conditionals and just set the locations if you want this cleaned up

pcdswidgets/vacuum/base.py
TrumanPar: I agree, I was struggling getting the dict scheme to work in a clean manner, I prefer a different solution. I'll get to work on it

pcdswidgets/vacuum/base.py
TrumanPar: Yes, I'll work on this as well

pcdswidgets/vacuum/base.py
ZLLentz: It's ok to skip nitpick comments if things run ok

pcdswidgets/vacuum/base.py
TrumanPar: I'll also investigate the failed unit test further, my first pass thought is the test is based on the assumption that there is only an icon in the layout, but now we're also adding the text box. But I'll take a deeper look this week

ZLLentz: Maybe the old test architecture is a bit flawed if it's so fragile. I leave it to you to decide whether to do the same thing again or do it differently, you don't have to fix the world in a single PR (or ever)

TrumanPar:
Pushing changes, the unit tests pass now that we don't initialize the textbox in __init__. However, I still need to create some new unit texts to confirm the textbox is good for future releases.

TrumanPar:
Superfluous commits aside, I think these unit tests are a good test of the text box's basic functionality. I'm also letting word wrap handle formatting the text, it seems to be working better now. All unit tests are passing

pcdswidgets/tests/vacuum/test_base.py
```diff
@@ -60,3 +60,45 @@ def test_icon_fixed_size(symbol):
...
+@pytest.mark.parametrize('location,layout,position',
+                         [(ContentLocation.Top, QVBoxLayout, 0),
+                          (ContentLocation.Bottom, QVBoxLayout, 1),
+                          (ContentLocation.Left, QVBoxLayout, 0),
+                          (ContentLocation.Right, QVBoxLayout, 1)],
+                         ids=['Top', 'Bottom', 'Left', 'Right'])
+def test_text_location(symbol, location, layout, position):
+    symbol.controlsLocation = ContentLocation.Bottom
+    symbol.channelsPrefix = "ca://area:function:device:01"
+    symbol.showName = True
```
ZLLentz, line 74: Maybe we've missed the case where `snowName` is `False` but that case is pretty simple so perhaps it doesn't urgently need a test

pcdswidgets/vacuum/base.py
```diff
@@ -197,6 +234,117 @@ def showIcon(self, value):
...
+        Override the textbox auto-generated from the channel prefix
+
+        Returns
+        -------
+        bool
+        """
+        if value != self._override:
+            self._override = value
+            if self._override and self.name is not None:
+                self.name.setText(self._override_name)
```
ZLLentz, line 293: Action item: is it possible for `self.name` to be `None` instead of `QLabel` when we set this?
Maybe if so we would skip this and apply it later when `self.name` is created or the layout is assembled.

pcdswidgets/vacuum/base.py
```diff
@@ -197,6 +234,117 @@ def showIcon(self, value):
...
+        Set the name when it is overriden
+
+        Returns
+        -------
+        str
+        """
+        if value != self._override_name:
+            self._override_name = value
+            if self._override and self.name is not None:
+                self.name.setText(self._override_name)
```
ZLLentz, line 320: Action item: same thought, is it possible for `self.name` to be `None`?

pcdswidgets/vacuum/base.py
```diff
@@ -197,6 +234,117 @@ def showIcon(self, value):
...
+        """
+        Set the font size for the name when rendering the widget.
+
+        Returns
+        -------
+        int
+        """
+        if value != self._font_size:
+            self._font_size = value
+            self.name.setStyleSheet(f"font-size: {self._font_size}px; background: transparent")
```
ZLLentz, line 346: Action item: same `None` comment as above

pcdswidgets/vacuum/base.py
```diff
@@ -197,6 +234,117 @@ def showIcon(self, value):
...
+        """
+        Whether or not to show the name when rendering the widget.
+
+        Returns
+        -------
+        bool (set to true will display the name)
+        """
+        if value != self._show_name:
+            self._show_name = value
+            if self.name is None:
```
ZLLentz, line 259: Food for thought: I think a lot of the awkward is-it-None error handling comes from delaying the creation of the widget. I know it's probably best practice to not make widgets that aren't used, but I'm at least a little bit worried about maintainability since we have to do a None check every time we want to access it.

pcdswidgets/vacuum/base.py
```diff
@@ -368,33 +516,69 @@ def assemble_layout(self):
...
-        }
-
-        layout = widget_map[self._controls_location][0]()
+        grouped_frame = QFrame(self)
+        grouped_widgets = QVBoxLayout()  # Default
+
+        # Determine what widgets to group
+        if self._text_location in [ContentLocation.Left, ContentLocation.Right] and self._text_location == self._controls_location:
+            grouped_widgets = QVBoxLayout()
+            if self.name:
```
ZLLentz, line 525: Nitpick: various python linters/style guides will suggest you use explicit `is None` or `is not None` checks whenever you expect something to be `None` because objects can evaluate to `False` sometimes (notorious examples including empty lists and blank strings). This isn't one of those cases so this is a nitpick rather than a full suggestion.
```suggestion
            if self.name is not None:
```

pcdswidgets/vacuum/base.py
```diff
@@ -368,33 +516,69 @@ def assemble_layout(self):
...
+                grouped_widgets.addWidget(self.name, alignment=Qt.AlignCenter)
+            grouped_widgets.addWidget(self.controls_frame, alignment=Qt.AlignCenter)
+
+            grouped_frame.setLayout(grouped_widgets)
+            layout_cls = QHBoxLayout
+            if self._controls_location == ContentLocation.Left:
```
ZLLentz, line 531: Comment: this rewrite is much easier to understand than the previous dict forest :+1:

pcdswidgets/vacuum/base.py
```diff
@@ -368,33 +516,69 @@ def assemble_layout(self):
...
+                widgets = [grouped_frame, self.icon]
+            elif self._controls_location == ContentLocation.Right:
+                widgets = [self.icon, grouped_frame]
+        elif self.name:
```
ZLLentz, line 535: Nitpick: same thought about `None` checks as above

pcdswidgets/vacuum/base.py
```diff
@@ -555,3 +739,20 @@ def update_status_tooltip(self):
         Set the tooltip on the symbol to the content of status_tooltip.
         """
         self.setToolTip(self.status_tooltip())
+
+    def format_name(self):
```
ZLLentz, line 743: Thing to know: nothing wrong with this function, but you can consider using the PyDM utilities for protocol/channel text: https://github.com/slaclab/pydm/blob/0806a50857d4de85ec5b88331a7f5b6ee55d9f4a/pydm/utilities/remove_protocol.py#L23

I don't think this is better than your implementation but I thought I'd point out that it exists

ZLLentz (commented):
One action item related to missing None checks in some of the property setters
The rest are really minor or are side thoughts, this is good

pcdswidgets/tests/vacuum/test_base.py
TrumanPar: I think that's a good idea. We might want to test that `icon` and `contentLocation` keep proper formatting with `name` initialized but hidden. The unit test would be similar to `test_controls_content_location` with small tweaks to account for the layout including `name`.

pcdswidgets/vacuum/base.py
TrumanPar: Agreed, I'll make the change

pcdswidgets/vacuum/base.py
TrumanPar: I'm personally on the side of initializing `name` until someone toggles `showName` in designer, but I'm at your mercy oh PR approver. I can initialize it like normal, and I would then have to edit the `test_controls_content_location` unit test to accommodate `name` in the layout and push again. Just let me know, and it will be done

pcdswidgets/vacuum/base.py
TrumanPar: This makes sense

pcdswidgets/vacuum/base.py
TrumanPar: good to know, thanks

pcdswidgets/vacuum/base.py
ZLLentz: Either way is ok and I'll approve either way so long as we check as needed

My opinion is that the added simplicity of always initializing name is worth the cost of having an extra object

pcdswidgets/vacuum/base.py
```diff
@@ -368,33 +514,69 @@ def assemble_layout(self):
...
+#            if self._controls_location == ContentLocation.Top:
+#                widgets = [self.controls_frame, self.icon]
+#            elif self._controls_location == ContentLocation.Bottom:
+#                widgets = [self.icon, self.controls_frame]
+#            elif self._controls_location == ContentLocation.Left:
+#                widgets = [self.controls_frame, self.icon]
+#            elif self._controls_location == ContentLocation.Right:
+#                widgets = [self.icon, self.controls_frame]
+#            else:
+#                widgets = [self.icon]
```
ZLLentz, line 567: If you're not planning to bring this back, it's time to delete these lines.

ZLLentz (dismissed):
I think you've done a good job here and have ended on a clean implementation. It's probably time to merge and defer further iteration until there's some feedback from users or other team members.

I left one last nitpick but I'm approving regardless :+1:

You can merge yourself as-is or you can handle the nitpick and I'll re-approve and merge afterwards

TrumanPar:
Got those lines removed, thanks for reviewing!

ZLLentz (approved):

ZLLentz:
@TrumanPar had you thought about when would be best to deploy these changes? I can pull them into the dev_conda any time (maybe now, maybe we wait for PAMM) and I'll tag this package so it ends up in the next pcds_conda tag.

TrumanPar:
@ZLLentz Let's do next Tuesday during the PAMM, would you want to deploy it around 2:30pm?

ZLLentz:
Sounds good to me!
Action items are something like:

- note the pre-update version
- git clone pull into the dev area
- spot check a bunch of vacuum screens
- claim victory, roll back, or tweak screens as appropriate

PR #97, opened 2026-04-23
SmarAct Widgets

aberges-SLAC:
## Description
Decided to pull some of the more useful SmarAct tip-tilt widgets into `pcdswidgets` for general use. I have a couple hang-ups for how I implemented a few things and would love suggestions/advice:
* The expert screen buttons, for now, will open a `SmaractOpenLoopContext` widget as a `Display`, which should be limited to the useful PVs for open-loop stages.
* Later on, I want this to call the expert pydm screen I'll eventually add to the next common release but that chicken needs an egg first.
* `SmaractOpenLoopContext` needed a class who subclassed `Display` for `PyDMRelatedDisplayButton` to work. I needed to do something tricky to get the `mro` to behave and pass the macros successfully. Unfortunately, the auto generation made this a designer aware widget and I think I did not want to do that with the `Display` subclass. 
    * The other widget is fine, because one could conceivably use that in a larger `ui` file.
   
I left `logger`s everywhere which was very helpful for `info` and `debug` prints when testing, but I removed most of my `info` prints to reduce terminal spam.

Some things I noticed trying this all out:
* The jinja template struggles with the macros in `PyDMRelatedDisplay`, Zach and I determined that the string property replacement should probably just use `'''` to handle all the silliness. 
* `PyDMRelatedDisplay` is a bit of a struggle. You need to do the `QTimer` method so you can have access to `_get_macros()` to pass it along to your display — _and_ you need to take special care to the `__init__` of the `Display` widget you're calling.
* You need set `icon=IconOptions.NONE` and add a `@staticmethod get_designer_icon() -> str:` function to tie into `pydm`'s hooks on designer if you want custom icons to show up.

I did not update the `README` with these discoveries, but I can if you'd like.

I also added a new widget sizing `Double` but did not add checks/tests for its enforcement. `Double` should be twice the height of `Full` at max, I'd say. 

## Motivation and Context
I've been wanting to overhaul the `SmarAct` GUI in the common ioc and get some lightweight screens out there and this seemed like a great place to add the basic components.

For closed-loop motors, I may keep the bulk of the `ui` in the common IOC itself, building off of other `pcdswidgets` components.

## How Has This Been Tested?
I have a test setup in the laser lab that I've been religiously testing against. See below for a brief video playing with the controls.

So far just here in this PR!

## Screenshots
image

image

Demo:

https://github.com/user-attachments/assets/7d068c98-0abe-49dc-9e6d-d81269819127

## Pre-merge Checklist
- [ ] Code works interactively


pcdswidgets/motion/smaract/smaract_open_loop_context_double.py
ZLLentz, line 1: Question: would it have been possible instead to just use the `.ui` file as the target for the related display button? Or do the classes here and their inclusion in designer have some additional purpose?

aberges-SLAC: I think I _have_ to do this to be test compliant, let me refactor and see what I can do. The context window itself for open-loop tip tilts may have some utility, but I personally am wary of the bloat. 

I think if I do it this way, I can avoid the second subclass and needed to do janky things with `Display` macros directly. Let me refactor and see if it helps.

aberges-SLAC:
@ZLLentz I am officially petitioning to reduce `ratio` to `0.70`, but the minimum dimension check is largely arbitrary either way _so_ I'm not sure what to do here.

I like keeping `Double` as 2x the size of `Full`. I could also write an exception into `Double`'s check but I'm not sure if I like that smell.

ZLLentz:
Yeah I think our options are:

- 0.7 is totally ok
- Is there a place for just having a "Large" size (with some good name) and not trying to police these?

aberges-SLAC:
> Yeah I think our options are:
> 
> * 0.7 is totally ok
> * Is there a place for just having a "Large" size (with some good name) and not trying to police these?

There could be, but it maybe also introduces/incentivizes folks naming their widget `Large` to avoid size constraints. IMHO it's not an easy line to toe. We want consistency in the GUI appearances but it also feels bad to police a constraint when/if the widget looks good as is (and performs to its intention).

tangkong:
cool designer icons 👍

tangkong:
The coverage test fails but the normal pip tests pass.  The flakiness is a bit tragic, perhaps there's some threading shenanigans to uncover

aberges-SLAC:
> The coverage test fails but the normal pip tests pass. The flakiness is a bit tragic, perhaps there's some threading shenanigans to uncover

I was just looking at this, perplexed to see it segfault in `SmarActTipTiltFull` but `Double` passed? Must be something fancy failing there. Somehow got an unbound method maybe?

pcdswidgets/generated/motion/smaract/smaract_open_loop_context_double_base.py
tangkong, line 1: My first thought was: should we gitignore these generated files?  If memory serves, all customization goes in other generated files (the ones not placed in the generated folder)

aberges-SLAC: I _think_ I was getting test failures on the environment when I excluded them, because the other designer-aware widgets depend on these modules existing for their classes to function.

So I think we unfortunately cannot, unless we have the test run `make` first?

tangkong: I think automatically running make on install is a reasonable adjustment. But outside the scope of this pr I imagine

aberges-SLAC:
> cool designer icons 👍

It made me sad that the context window didn't have a super special icon, so I fixed that and made this PR slightly more annoying.

pcdswidgets/generated/motion/smaract/smaract_open_loop_context_double_base.py
ZLLentz: I've intentionally abandoned an old version of the code that runs `make` on install because it increased complexity in lots of ways and had annoying interactions with the dev workflow here- you'd `make` which did a `pip install` which called `make`, etc.

ZLLentz:
I did just finish a PR to make icons easier, but let's merge this one first if we can

pcdswidgets/generated/motion/smaract/smaract_open_loop_context_double_base.py
tangkong: I see.  I rescind my comment 🫡

aberges-SLAC:
Just waiting on the approval then I'll gladly merge [:

pcdswidgets/motion/smaract/smaract_tip_tilt_double.py
```diff
@@ -0,0 +1,137 @@
...
+
+        motor_pv = self._get_macro(f"{axis}_motor")
+
+        if not motor_pv:
+            logger.debug(f"Macro for {axis}_motor does not yet exist")
+            return
+
+        button = getattr(self, f"{axis}_expert_screen")
+
+        button.setFilenames([str(Path(__file__).parents[2] / "ui/motion/smaract/smaract_open_loop_context_double.ui")])
```
ZLLentz, line 90: Two thoughts: one action item question and one existential musing

1. Do these ui files get bundled in the python install? I'm worried that they might not, and that this wouldn't work properly once we install the library and don't have the git clone handy.
2. I wonder if this reveals a hole in the widgets paradigm in this module- you'd naively just want to include a path to a screen as your expert helper but that screen is no longer at a fixed location, so including it in your ui file seems intractable, leading to dynamic path discovery here. This causes clear friction for both related displays and embedded displays and warrants some investigation from me.

pcdswidgets/motion/smaract/smaract_tip_tilt_double.py
```diff
@@ -0,0 +1,137 @@
...
+    @staticmethod
+    def get_designer_icon() -> str:
+        """Icon for usage in Qt designer."""
+        return QIcon(str(Path(__file__).parent / "smaract_tip_tilt_qt_icon.png"))
+
+    def _setup_expert_screens(self):
+        """Macros aren't immediately available through _get_macros, wait until they are."""
+        if not self._get_macro("vertical_motor") or not self._get_macro("horizontal_motor"):
+            self._macros_timer.start()
+            return
```
ZLLentz, line 61: Throwaway comment: there was probably some way to do this without the timers but I cba to figure it out so I'm also not going to ask anyone else to figure it out.

Would it be helpful to be able to add macro callbacks as a library feature on `DesignerWidget`? Or is this not worth the brainspace?

E.g. something like this to be put in the init
```
self.add_macro_callback(self.my_func, "vertical_motor")
```
could call my_func with the value of the "vertical_motor" macro once ready

pcdswidgets/motion/smaract/smaract_tip_tilt_full.py
ZLLentz, line 1: How similar are these two files? Is there any sense in consolidating the logic, or is it not worth the effort?

pcdswidgets/motion/smaract/smaract_tip_tilt_double.py
aberges-SLAC: 1. I think that since they get auto-generated, they'd have to somewhere, right? But honestly, I didn't check for this and it is probably worth it to see. 
2. This here is the real worry for me — without going the full custom settings panel class route that I went with `qmini` in `pcdsdevices`, there isn't a good way to open _just_ a `py` as a display without explicitly subclassing `display`. 

I'm hoping this selfishly goes away for SmarAct when I can use `pcdswidgets` down the line to put expert screens in the common IOC in a static way, but it is a real 🐔 and 🥚 disaster

ZLLentz:
I _think_ the only actionable thing in my review that gives me pause is not knowing if the .ui is bundled in the install

But I also think all of my comments are in the "can do this later" camp, so I'm going to leave my approval now

ZLLentz (approved):
Looking good, let's get some widgets in and around!

pcdswidgets/motion/smaract/smaract_tip_tilt_double.py
ZLLentz: right- expert screens in the common IOC (a fixed path) does alleviate a lot of these issues

ZLLentz:
(on the side, I'm going to figure out how to check this .ui file bundling concern)

pcdswidgets/motion/smaract/smaract_tip_tilt_double.py
aberges-SLAC: Callback that fires once is probably the equivalent, but I honestly could've think of any other way to do this without `QTimer` or some more `pcdswidgets` infrastructure change and I was already creeping the scope.

The more I think about this, the more I like it because it enables some clever updating when the macro changes which might be useful for some screens. The only use-case on my mind is this exact tip-tilt widget with the button inversion, but I have that solved the `pydm` way anyway and maybe it is better to not propagate the macro changing throughout the widget.

pcdswidgets/motion/smaract/smaract_tip_tilt_full.py
aberges-SLAC: They are 99.9, if not 100% identical as they exist right now.

There might be a future where `SmarActTipTiltDouble` is actually aware of whether or not is encoded, and may want `PyDMPushbButton`s that exist for the annoying home+calibrate routine that these closed loop stages want. Then I open Pandora's box of `QThread` management that I did in `pcdsdevices`, but I don't really see that going away. Thinking about it more, that's extension and not replacement so I don't really see why I wouldn't just _use_ the majority of this code anyway.

Would the approach be making a `smaract_tip_tilt_mixins.py` package in `pcdswidgets/motion/smaract`, then just having both of the designer widgets subclass it? I think that's what I did with `pcdsdevices.qmini` but I want avoid naming gore by having `SmaractTipTiltDouble(SmaractTipTiltDoubleBase, SmarActTipTiltBase)`, but maybe that isn't as bad as I'm making it out to be in my head.

ZLLentz:
Yeah it bundles ok
```
/cds/home/z/zlentz/temp/widget_venv_test/lib/python3.12/site-packages/pcdswidgets/ui/motion/smaract/smaract_open_loop_context_double.ui
```
Sometimes these are set up to not include data files- but this repo seems fine

ZLLentz:
I think it must be this thing:
https://github.com/pcdshub/pcdswidgets/blob/a570581e44fd1efb77cbc6b3b3bb62da061ff77c/pyproject.toml#L22
I'm going to do a few minutes of reading the docs because maybe I can simplify the other PR where I tried extra to include the png

ZLLentz:
OK, the trick is that data files _inside_ the module are installed, as long as they are tracked in your vcs. The extra effort for `MANIFEST.in` in mainly for untracked files and files outside the module that you want to distribute with your package.

Why's it gotta be confusing

pcdswidgets/motion/smaract/smaract_tip_tilt_full.py
ZLLentz: Yeah, either shared mix-ins or importable helper functions that encapsulate the code- either may be more convenient in some situations

PR #98, opened 2026-04-24
API: icons refactor

ZLLentz:
## Description

- Move the old "icons" code to the "symbols" module to avoid the obvious name collisions that come with using "icons" to mean "not icons"
- Use the "icons" folder as a place to store designer icons, and seed it with a single tiny test icon
  - Move Adam's icons into this folder too
- Allow using a `str` as the designer options icon argument to load from the icons folder
- Update the readme appropriately
- Consolidate the icon generation code in a more convenient, readable way

## Motivation and Context

- This was documented wrong before and it was a pain for Adam and I figure it out in #97 
- This should be easier for everyone to use

## How Has This Been Tested?

Interactively using the test icon
I added some "smoke"-level unit tests

In the README

## Screenshots

Here's what it looks like when you include a png icon:

image

## Pre-merge Checklist


ZLLentz:
I'm going to resolve conflicts with #97 and do the obvious simplifications that are enabled by this PR specifically.

pcdswidgets/builder/designer_widget.py
```diff
@@ -2,18 +2,24 @@
...
 from pydm.widgets.qtplugin_extensions import RulesExtension
+from qtpy.QtGui import QIcon
 from qtpy.QtWidgets import QAction, QDialog, QFormLayout, QHBoxLayout, QLineEdit, QPushButton, QVBoxLayout, QWidget
 
+import pcdswidgets
+
 from .designer_options import DesignerOptions
 from .icon_options import IconOptions
 
+ICON_AREA = Path(pcdswidgets.__file__).parent / "icons"
```
tangkong, line 21: Could you not just navigate to the icons directory from this file?  

Of course if the pcdswidgets import isn't heavy this isn't an issue but we've been burned by this before

tangkong (dismissed):
This looks good to me on the whole, I can't argue with those who have battle tested it.

pcdswidgets/builder/designer_widget.py
ZLLentz: The main reason I've switched to relative from top paths is because I've been burned a few times from relative filepaths when I move my module files around.

This should be a separate concern than the python code imports.

ZLLentz:
Adam- I'm adding you as reviewer here because you were looking at icons and I ended up tweaking your files to load icons via the icons folder

aberges-SLAC:
@ZLLentz This looks great to me! Glad to see it incorporated as the default option when given a `str` path, should make this a lot cleaner to handle in the future ~~when I make many more custom icons~~

aberges-SLAC (approved):

PR #100, opened 2026-05-15
Add "Fixed" and "Stretch" Widget Size Classes

ipotanin:
## Description

Modified the README and CI tests that enforce widget size to classes. Added two size classes.
- Fixed -> defines a custom widget size explicitly in the name  (enforces min and max size)
   -  example motor_classic_fixed_400X400.ui
- Stretch class -> define a widget designed to be expanded to fill a screen (only enforces min size)
  -  example motor_classic_stretch_400X400.ui
## Motivation and Context

Following an offline conversation with "ZLLentz", there was some discussion around loosening the size restriction on widgets as long as the size was explicitly defined. This PR is my suggestion for this.

## How Has This Been Tested?

built some widgets with these size classes and made sure the test failed if sizes were not correct

here
## Screenshots

NA

## Pre-merge Checklist
- [X] Code works interactively
- [X] Code contains descriptive docstrings, including context and API
- [X] Test suite passes locally
- [X] Test suite passes on GitHub Actions

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,8 +103,7 @@ Before getting too deep, however, please consider widget sizing:
 ### Widget Sizing
 We have some strict guidelines on widget sizing. These are established to give us some consistency in application of widgets, as well as to make it simpler to avoid resizing a widget between library releases.
 
-Device control widgets should fall into exactly one of three size classes.
-(Note: we can add more size classes if necessary).
+Device control widgets should fall into exactly one of the size classes below.
 
 To ensure sizing consistency, set the minimum and maximum sizes to values that look good throughout the range
 and are permissible sizes as recorded below.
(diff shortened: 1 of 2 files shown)


ipotanin:
@ZLLentz 
I wanted to get you opinion on this proposed change to the widget size policy.

ZLLentz:
I think it's reasonable, but what was explicitly asked of me offline was to remove the size classes altogether. So I think we should do that instead (https://jira.slac.stanford.edu/browse/ECS-10450). We may even consider not policing it at all.

I wonder if having the size in the name dooms us even more: we would want to rename the widget to resize it, which completely breaks all the screens immediately because the old name is gone.

I still think suffix names are super helpful for conveying intent- finding the "Row" widget for example is helpful to have in the name. In that regard: I think "stretch" is a great suffix.

PR #101, opened 2026-05-19
SVG Multi State LED Widget

aberges-SLAC:
## Description
As discussed in ecs-gui chat, I have finally gotten around to making a multi-state SVG widget based on PyDMSymbol that combines both color and shape changes for different states in .MSTA.

There is now:
* `SvgMultiStateLED` widget, which defaults to `${motor}.MSTA` as the main channel to evaluate. 
* `icons.glyphs.py` module which is just a collection of paths as constants which should help with relative import nonsense

There is included functionality for additional channels, but I can be easily convinced to forgo this for this PR.

## Motivation and Context
Laser goggles are apparently _too_ good at attenuating certain colors, making circle LEDs somewhat useless when convey state changes to a user. Combining both color and silhouette changes to indicate the present state, we get the best of both worlds. Or at least the mid of both.

This comes with a price: centering the widget made me include the `PyDMSymbol.paintEvent()` method as an overwrite with barely any changes, much to my chagrin. Also, if you want to change the color _outside_ of the definition in svg good luck. The paint event relies on it so you need to load the whole code into memory and edit it there, dynamically, if you wanted to do this.

Just make new svgs.

## How Has This Been Tested?
Tested this on `LAS:LLN:MCS2:01:m1` and had some decent results.

Here and #ecs-gui!

## Screenshots

https://github.com/user-attachments/assets/3ff2408f-ff66-4206-9d9d-55a553ca8940

## Pre-merge Checklist
- [ ] Code works interactively


ZLLentz:
The first response a developer has to that video is "looks snazzy"
The first response a user might have is "why is the arrow pointing right when I move left?"

Good job, I'll peek at the code

ZLLentz:
Regarding the checks failure- I'm probably just going to remove the size checks from the repo before deciding what to replace it with

pcdswidgets/motion/common/svg_multi_state_led.py
```diff
@@ -0,0 +1,320 @@
...
+        Connect init_channel. This assumes the MSTA motor field and CA.
+        Change or overwrite this method if you want to begin with a different
+        PV.
+        """
+        if not self._motor:
+            # Haven't set the property yet
+            self._macros_timer.start()
+            return
+
+        self.set_channel(f"ca://{self._motor}.MSTA")
```
ZLLentz: Suggestion: can we forgo the timer and do this in `set_motor` instead?

I think this would be cleaner and it would also allow us to set a different motor either at runtime or in designer rather than taking the first one set forever.

pcdswidgets/motion/common/svg_multi_state_led.py
```diff
@@ -0,0 +1,320 @@
...
+                scale = (sf, sf)
+            self._painter.scale(scale[0], scale[1])
+            self._painter.drawPixmap(event.rect().x(), event.rect().y(), image_to_draw)
+        elif isinstance(image_to_draw, QSvgRenderer):
+            draw_size = QSizeF(image_to_draw.defaultSize())
+            draw_size.scale(QSizeF(event.rect().size()), self._aspect_ratio_mode)
+
+            # CENTER THE SVG, so annoying this had to be done.
+            x = (event.rect().width() - draw_size.width()) / 2.0
+            y = (event.rect().height() - draw_size.height()) / 2.0
```
ZLLentz, line 110: Question: Is this the main/only thing you changed in `paintEvent`? I'm considering whether I ought to add this to the list of things-to-add-as-options-upstream "eventually".

pcdswidgets/motion/common/svg_multi_state_led.py
```diff
@@ -0,0 +1,320 @@
...
+        Override to force disconnected icon when not connected.
+        This is called by PyDMWidget regularly.
+        """
+        if not hasattr(self, "_state_dict"):
+            logger.debug("_state_dict does not exist!")
+            super().check_enable_state()
+            return
+
+        # Check parent's connection state
+        if hasattr(self, "_connected") and not self._connected:
```
ZLLentz: Suggestion: you can save yourself some implementation grief by setting every attribute you use to a default value in `__init__`. For example, where you set the starting state to "disconnected" you could also do `self._connected = False` and then you don't have to worry about attribute errors later.

pcdswidgets/motion/common/svg_multi_state_led.py
```diff
@@ -0,0 +1,320 @@
...
+            logger.debug(f"Motor {self._motor} disconnected")
+        else:
+            logger.debug(f"Motor {self._motor} connected")
+
+    def check_enable_state(self) -> None:
+        """
+        Override to force disconnected icon when not connected.
+        This is called by PyDMWidget regularly.
+        """
+        if not hasattr(self, "_state_dict"):
```
ZLLentz: Question: is it possible for this to not exist? It's set pretty explicitly in `__init__`.

pcdswidgets/motion/common/svg_multi_state_led.py
```diff
@@ -0,0 +1,320 @@
...
+                ch_obj = PyDMChannel(
+                    address=channel_addr,
+                    value_slot=lambda val, i=idx: self._additional_value_callback(i, val),
+                    severity_slot=lambda sev, i=idx: self._additional_severity_callback(i, sev),
+                    connection_slot=lambda conn, i=idx: self._additional_connection_callback(i, conn),
+                )
+                ch_obj.connect()
+                self._additional_channel_objs.append(ch_obj)
+
+    def _additional_value_callback(self, idx: int, value: any) -> None:
```
ZLLentz: Nitpick comment:

I think you mentioned something about this in the PR description: I assume the intention is to extend these in subclasses?

The reason I don't like these isn't because they don't do anything yet, but rather because they are exposed to `designer` and don't do anything yet. I don't like leaving a trap to confuse fellow team members, someone might use the property, observe no behavior, and then ask about it.

Anyway, I think maybe this should be separated into another widget that uses them by default- maybe one that is tuned to "show the worst alarm" instead of "show the msta" for non-motor stuff.

carogee:
Not specifically about the code, but did you have a chance to test visibility of all the icons with various laser goggles?

pcdswidgets/motion/common/svg_multi_state_led.py
aberges-SLAC: I think I tried this and it didn't work appropriately (tries to set an empty field before it receives the motor value). But I started this before we had the other changes so maybe I am misremembering. Let me try this, I would love to abandon the timer.

pcdswidgets/motion/common/svg_multi_state_led.py
aberges-SLAC: Yes and it makes me very very sad.

pcdswidgets/motion/common/svg_multi_state_led.py
aberges-SLAC: Good point, this is probably a consequence of a lot of redundancy I added yesterday during bug fixing and troubleshooting. I should probably give this a full pass and see what I should simplify before I push this PR any further

pcdswidgets/motion/common/svg_multi_state_led.py
aberges-SLAC: Surprisingly, yes. On start-up the dict is called but doesn't exist. I only found out during debug messages in the logger. It _might_ honestly be ignorable. But I think since then I moved the state dict before the `super().__init__` call, which might have prevented this.

pcdswidgets/motion/common/svg_multi_state_led.py
aberges-SLAC: Yeah I was definitely leaning this way, I have a specific SmarAct implementation in mind and it probably warrants just subclassing to that specific example.

I'm gonna opt to remove the additional channel functionality

aberges-SLAC:
> Not specifically about the code, but did you have a chance to test visibility of all the icons with various laser goggles?

Not yet! But that's a very good point. There's a few in RIX I can use, but there are probably too many flavors of laser goggles across the hutches

aberges-SLAC:
> The first response a developer has to that video is "looks snazzy" The first response a user might have is "why is the arrow pointing right when I move left?"

Hahaha I thought the same thing but honestly the overhead required for "moving left" and "moving right" as distinct states and SVGs to call, it felt silly to force in. I might do it anyway if user feedback demands it

ZLLentz:
> the overhead required for "moving left" and "moving right" as distinct states

Is this more complicated than checking the first bit in MSTA? Or is that so raw that it doesn't invert itself when we invert the motor in EPICS?

ZLLentz (approved):

aberges-SLAC:
> > the overhead required for "moving left" and "moving right" as distinct states
> 
> Is this more complicated than checking the first bit in MSTA? Or is that so raw that it doesn't invert itself when we invert the motor in EPICS?

It's more of, you don't know if left is actually left from software level. `MOTOR.RECORD` only knows direction in terms of forward/backward, but for some deployments this is inverted based on the design (I'm singling out tip-tilt devices specifically). So for linear motors I think this is fine, as long as forward is always "right" — but that ultimately depends on physical implementation and whatnot.

Not to mention my little trick in `SmarActTipTilt` widgets for inverting the axes (which some users requested because periscopes be confusing sometimes) wouldn't change this MSTA bit.

There's enough exceptions where I think this could be messy to implement as opposed to just "arrow always mean it move"

PR #104, opened 2026-05-28
Modify layout and fonts for motor_classic widgets and add motor_slit widget

carogee:
## Description

Improved layout and standardized fonts/sizing for the motor_classic_row and motor_classic_full widgets. Also added a widget for Beckhoff motor slits.

## Motivation and Context

Motor classic widget layout and fonts were previously not standardized.

## How Has This Been Tested?

Tested with XCS and XPP motors.

## Screenshots

designer

motor_classic
slits
BeckhoffSlits

## Pre-merge Checklist


pcdswidgets/ui/motion/common/motor_classic_full.ui
```diff
@@ -358,30 +608,70 @@
...
+          </property>
+          <property name="PyDMToolTip">
+           <string/>
+          </property>
+          <property name="monitorDisp">
+           <bool>false</bool>
           </property>
           <property name="channel">
-           <string>ca://${MOTOR}.STOP</string>
+           <string>ca://${MOTOR}:SEQ_SELN</string>
```
ZLLentz, line 638: Question/suggestion: does `SEQ_SELN` exist in anything other than the `ims` IOC? If it's `ims`-only, I don't think it belongs in the generic motor widget.

What exactly is it for? Is it the error reset maybe?

There might be a simple way to make variants of this widget sourced from the same layout that substitute in different error reset PVs.

pcdswidgets/ui/motion/common/motor_beckhoff_slits.ui
ZLLentz, line 1: Question: where did the built file at `pcdswidgets/motion/common` related to the slits go? It's not in this PR.
Change request: add this file so that the widget is included!

pcdswidgets/ui/motion/common/motor_beckhoff_slits.ui
```diff
@@ -0,0 +1,966 @@
...
+   <widget class="QWidget" name="layoutWidget">
+    <property name="geometry">
+     <rect>
+      <x>10</x>
+      <y>10</y>
+      <width>781</width>
+      <height>91</height>
+     </rect>
+    </property>
+    <layout class="QGridLayout" name="gridLayout">
```
ZLLentz, line 46: Question/suggestion: the grid layout here is a bit strange- there's a middle row that exists but has no widgets in it. Was this intentional to make the spacing bigger between the rows? Was there no built-in spacing option that could have been tweaked instead?

ZLLentz (commented):
I'm amused that copilot noticed the missing file

I had a few things to point out, nothing major but they need consideration

pcdswidgets/ui/motion/common/motor_beckhoff_slits.ui
carogee: added

pcdswidgets/ui/motion/common/motor_beckhoff_slits.ui
ZLLentz: An amendment: I see now that there's an intentional vertical expander in the middle

carogee: I added a vertical spacer in the middle row to create extra space. I did not find another vertical spacing method, but can look if you know of one.

ZLLentz: I hadn't noticed the vertical expander on last review, but there is a built-in layout option for fixed-size spacing between rows

pcdswidgets/ui/motion/common/motor_classic_full.ui
carogee: I only see it in the IMS motors, but for the Beckhoff motors it is .bReset. I'm not seeing anything in the epics motor record that is a universal reset or clear. Maybe STOP or SPMG is the closest?

ZLLentz: Yeah, motor record doesn't have a specification for clearing errors and every IOC does it differently. I don't think there are any generic substitutes.

We need an error reset button though, right?

So, we should probably write some code in https://github.com/pcdshub/pcdswidgets/blob/master/pcdswidgets/motion/common/motor_classic_full.py to help us pick the right variant depending on the stage type.

I'm thinking either adding a property that lets you pick Beckhoff, IMS, etc. or maybe just multiple classes that use the same base class but one for each type and then we fill in the clear error PV accordingly, or hide it if that motor type doesn't have a way to do it.

pcdswidgets/ui/motion/common/motor_beckhoff_slits.ui
carogee: Ah thanks, that is definitely a cleaner way to set the spacing!

ZLLentz:
@carogee do you see yourself having time to give different error reset PVs for different motor types a go? Or should I give it a try in a follow-up PR?

ZLLentz:
(I hadn't realized Carolyn is out of the office- I'll decide what to do with this next week because I want it in)

ZLLentz:
I'm intending to:
- Review this one more time
- Resolve merge conflicts
- Merge
- Make a follow-up PR to add infrastructure for selecting which motor type we're using (to get the correct error reset button), starting with only having an IMS option.

pyproject.toml
```diff
@@ -42,6 +42,8 @@ GetterPump = "pcdswidgets.vacuum.pumps:GetterPump"
 HotCathodeComboGauge = "pcdswidgets.vacuum.gauges:HotCathodeComboGauge"
 HotCathodeGauge = "pcdswidgets.vacuum.gauges:HotCathodeGauge"
 IonPump = "pcdswidgets.vacuum.pumps:IonPump"
+MotorBeckhoffRow = "pcdswidgets.motion.common.motor_beckhoff_row:MotorBeckhoffRow"
```
ZLLentz, line 45: This doesn't actually exist, I'm confused how it ended up here and how this didn't break in testing. I'll need to remove this line.

ZLLentz:
```suggestion
```

ZLLentz: But I'll do this in my follow-up PR I think

uv.lock
ZLLentz, line 1: This is the file that breaks the merge. Since this PR is from Carolyn's master branch, I'm not sure if I can resolve this easily, I may have to make a full branch copy and a second PR to complete the merge.

ZLLentz: The regressions and changes here imply a non-standard dev environment was used I think?

ZLLentz (approved):
Approving, but closing this in favor of merging #113, which is the same diff except I merged into master and solved the merge commits. The original commits are preserved.

PR #105, opened 2026-05-30
Feature finder

aberges-SLAC:
## Description
This is a port of https://github.com/pcdshub/lcls-pydm-common/tree/master/1d_scan into `pcdswidgets` where it is better served. This generic "feature finder" widget couples a standard motor record PV to any read-access PV that may be coupled or optimized by said motor. 

The motor behavior is limited and does not account for retries or IMS chicanery. So far, I have only tested this with SmarAct closed loop motors and behaves accordingly.

## Motivation and Context
Scientists requested a generic "scanner" application to optimize their signal when searching for THz generation in qRIX. This app ended up being generic enough that supporting it as a widget seemed optimal.

Extra feature requests were added upon this conversion:
* Save the settings at the _user_ level for a specific PV combination
* Allow the user to change the PVs after the widget has been launched. 
    * Although I tried to make this safe, this is a use-at-your-own-peril type ordeal. At the very least you should not be moving multiple motors

## How Has This Been Tested?
I tested this on my test SmarActs in the laser lab, `LAS:LLN:MCS2:01:m1` which is encoded linear stage. See attachments.

This code, PR, and here: https://jira.slac.stanford.edu/browse/ECS-10524

## Screenshots
image

image

image

https://github.com/user-attachments/assets/bae89d82-8219-4765-8aa0-28eaafa57861

## Pre-merge Checklist


pcdswidgets/icons/feature_finder_icon.png
KaushikMalapati, line 1: Is this a radar?

KaushikMalapati:
The name feature finder seems a little generic imo. Feel like it should at least have scan in the name but it's fine if the people using it already know it as feature finder.

pcdswidgets/common/tools/feature_finder.py
```diff
@@ -0,0 +1,886 @@
...
+        self._macros_timer.setSingleShot(True)
+        self._macros_timer.start()
+
+    def post_init_setup(self) -> None:
+        """
+        Need to wait for the object to exist for its macros to be available
+        """
+        if not self._get_macro("motor") or not self._get_macro("detector"):
+            self._macros_timer.start()
+            return
```
KaushikMalapati: Seems a little odd to do it this way. If you change the motor or detector, we don't re-run post_init on the new configuration. I might have instead of overridden _set_macro and run the rest of post_init_setup if motor and detector are defined and either of them are being set.

pcdswidgets/common/tools/feature_finder.py
```diff
@@ -0,0 +1,886 @@
...
+        except Exception as e:
+            logger.warning(f"Failed to load settings: {e}")
+
+    def connect_motor(self) -> None:
+        """
+        Once the motor macro is available, connect all relevant attr, widgets, etc.
+        """
+        self._motor_pv = self._get_macro("motor")
+        # Get the EGU as a channel
+        self.motor_egu_get.hide()
```
KaushikMalapati: Instead of having a pydm widget you hide in order to interact with a PV, you can create a [PyDMChannel](https://github.com/slaclab/pydm/blob/57fb46b7f1f78700b92f01a04de098f05b6ef1e3/pydm/widgets/channel.py#L18). Here is an example 
https://github.com/pcdshub/ioc-common-qadc/blob/d3d4d7d5d91e0ed1eac0311aa4eb5094ca3cb85b/qadcScreens/widgets.py#L222-L223
This is what all the widgets use under the hood anyways

pcdswidgets/common/tools/feature_finder.py
```diff
@@ -0,0 +1,886 @@
...
+        """
+        Wait for widgets to actually connect and get new data, then refresh.
+        """
+        widget: QDoubleSpinBox
+
+        # We'll use the EGU widgets as our litmus
+        if (not self._motor_egu or not self._detector_egu) and retry < 20:
+            # Give up after 2 seconds
+            QTimer.singleShot(100, lambda: self.finish_update_pvs(retry + 1))
+            return
```
KaushikMalapati: This feels a little odd to me too. Maybe we can use the connection_slot  of a pydmchannel instead of retrying for twenty seconds?

KaushikMalapati (dismissed):
Left a few nits but nothing that should stop this from being merged. I mainly looked at a diff between feature_finder.FeatrureFinder and 1d_scan.SimpleScannger and just skimmed over the rest which was already reviewed a few days ago.

ZLLentz:
> Feel like it should at least have scan in the name

It's kind of funny that omitting scan from the name was the _only_ requirement given when we were asked to rename this, to avoid confusion with e.g. DAQ scans and similar

pcdswidgets/common/tools/feature_finder.py
```diff
@@ -0,0 +1,886 @@
...
+    upper_limit_get: QLabel
+    bwd_run: PyDMPushButton
+    bwd_step: PyDMPushButton
+    stop: PyDMPushButton
+    fwd_run: PyDMPushButton
+    fwd_step: PyDMPushButton
+    reset_graph: PyDMPushButton
+    invert_x_checkbox: QCheckBox
+    plot: PyDMScatterPlot
+    change_pvs_button: PyDMPushButton
```
ZLLentz: Note to self: need to revise the auto-type-hints in the base class to include all the widgets so we don't have to place them redundantly here

pcdswidgets/common/tools/feature_finder.py
```diff
@@ -0,0 +1,886 @@
...
+
+        # Data arrays for plotting
+        self._motor_positions = []
+        self._detector_values = []
+        # detector
+        self._detector_pv = ""
+        self._detector_egu = ""
+
+        # Config file for caching
+        self._config_file = Path.home() / ".config" / "feature_finder_settings.json"
```
ZLLentz, line 73: Comment: I gave this some renewed review consideration now that this is a config file for a widget instead of a standalone application, the format chosen for the application is still appropriate for the widget since it has a separate entry for each motor/detector pair

pcdswidgets/common/tools/feature_finder.py
```diff
@@ -0,0 +1,886 @@
...
+
+        # Cross-hairs
+        self._crosshair_timer = QTimer()
+        self._crosshair_timer.timeout.connect(self._update_crosshair)
+        self._crosshair_timer.start(100)  # Update every 100ms
+
+        # Some stylization and housekeeping on QLabels
+        ## Step size
+        self.step_size_set.setMinimum(-inf)
+        self.step_size_set.setMaximum(inf)
```
ZLLentz, line 97: Comment: does qt no longer yell (stderr) at you for making these too big? Or am I remembering something wrong?

pcdswidgets/common/tools/feature_finder.py
ZLLentz: I also would have hooked into the macro setting behavior instead of doing a one-time QTimer

pcdswidgets/common/tools/feature_finder.py
ZLLentz: This is done correctly a few lines below for the dmov

pcdswidgets/common/tools/feature_finder.py
```diff
@@ -0,0 +1,886 @@
...
+        Generate a unique key for this motor/detector pairing.
+
+        Returns
+        -------
+        str
+            Configuration key in format 'motor_pv::detector_pv'
+        """
+        return f"{self._motor_pv}::{self._detector_pv}"
+
+    def save_settings(self) -> None:
```
ZLLentz, line 176: Comment: I double-checked that this is called at all the right times and it is :+1:

ZLLentz (dismissed):
Comments are all minor or in agreement, sending my approval

pcdswidgets/icons/feature_finder_icon.png
aberges-SLAC: [; yes

aberges-SLAC:
> The name feature finder seems a little generic imo. Feel like it should at least have scan in the name but it's fine if the people using it already know it as feature finder.

Funny enough, we decided on this name to explicitly deconflict it with the myriad of "scanner" GUIs cropping up

pcdswidgets/common/tools/feature_finder.py
aberges-SLAC: Perhaps it is much cleaner to override `_set_macro`. Perhaps that would actually make the PV changing feature less of a code-bloat implementation.

Let me tinker with this, as I already had to re-invent the wheel a bit because `_set_macro` doesn't call `set_channel` on widgets with those `attr` and maybe I should employ it there for brevity.

pcdswidgets/common/tools/feature_finder.py
aberges-SLAC: Except there is no way to force a standalone `PyDMChannel` to process or `get()`, it'll only update on the next `monitor` event. As `.EGU` pretty much never change, this was the cleanest way I could do it.

I initially tried with just a `PyDMChannel` and I couldn't get it to behave the way I needed it to. If there is a trick to this I'd love to know!

pcdswidgets/common/tools/feature_finder.py
aberges-SLAC: It retries for a maximum of 2 seconds, but in practice never it makes it past 2-3 retries. The problem I have is that `self.motor_egu_get.connected` is indeed `True` but the cache is stale and reading the previous value. Significantly less noticeable than when this happens with `self.detector_egu_get`, which ends up writing the incorrect EGU on the plot. 

This one was an annoying bug to tackle and I'm personally not happy with the timer solution as well

pcdswidgets/common/tools/feature_finder.py
aberges-SLAC: I haven't gotten a single complaint from qt about this, surprisingly. But I also think I am using the max value that Qt expects anyway so it doesn't care (`inf` is actually just `max` `float`)

ZLLentz: Ah, maybe the `QSpinBox` is the one that yelled at me, rather than `QDoubleSpinBox`

pcdswidgets/common/tools/feature_finder.py
ZLLentz: Maybe it's not worth fussing over, but I would have expected to either be able to get the value here with a channel or check the units that get bundled with the setter widget. Maybe I should tinker a bit.

KaushikMalapati: The widget's value is also updated on monitor events through a value callback so it should work the same way, except you have to define your own callback

aberges-SLAC: I think even in the `motor_is_moving` callback implementation with the `.DMOV` `PyDMChannel`, I don't see any processing until the monitor event posts. So unless `EGU` changes I don't get a value, I just get `None` or a stale cache.

ZLLentz: There are definitely some inconsistencies in how pydm populates non-updating signals, but I'm seeing at least the value slot consistently getting a "first" value that can be used in a callback. It can't be used synchronously like Adam is doing here though (but I suspect this implementation must have a rare race condition because of how pydm works?)

See my testing where I expected most of these to work but only a few of them did:

<details>

```
zlentz@psbuild-rocky9-01:~$ cat test_egu.py
from os import path

from pydm import Display
from pydm.widgets.channel import PyDMChannel
from PyQt5.QtCore import QTimer


class TestEgu(Display):

    def __init__(self, parent=None, args=None, macros=None):
        super().__init__(parent=parent, args=args, macros=None)
        self.egu_ch = PyDMChannel("ca://LAS:LLN:MCS2:01:m1.EGU", value_slot=self.show_egu_raw)
        self.egu_ch.connect()
        self.rbv_ch = PyDMChannel("ca://LAS:LLN:MC2:01:m1.RBV", unit_slot=self.show_egu_meta_rbv)
        self.rbv_ch.connect()
        self.main_ch = PyDMChannel("ca://LAS:LLN:MC2:01:m1", unit_slot=self.show_egu_meta_main)
        self.main_ch.connect()
        print("show_egu_label", self.val_label._unit)
        self.timer = QTimer()
        self.timer.setInterval(1000)
        self.timer.timeout.connect(self.show_egu_label)
        self.timer.start()

    def show_egu_raw(self, *args, **kwargs):
        print("show_egu_raw", args, kwargs)

    def show_egu_meta_rbv(self, *args, **kwargs):
        print("show_egu_meta_rbv", args, kwargs)

    def show_egu_meta_main(self, *args, **kwargs):
        print("show_egu_meta_main", args, kwargs)

    def show_egu_label(self, *args, **kwargs):
        print("show_egu_label", self.val_label._unit, args, kwargs)

    def ui_filename(self):
        return "test_egu.ui"

    def ui_filepath(self):
        return path.join(path.dirname(path.realpath(__file__)), self.ui_filename())

zlentz@psbuild-rocky9-01:~$ pydm test_egu.py
[2026-06-01 13:46:33,345] [INFO    ] - Using PyDM via SSH. Reverting to Software Rendering.
show_egu_label
show_egu_raw ('mm',) {}
show_egu_label mm () {}
show_egu_label mm () {}
show_egu_label mm () {}
show_egu_label mm () {}
show_egu_label mm () {}
show_egu_label mm () {}
```

</details>

aberges-SLAC: I'm glad this was consistent with my testing in the widget! I was pulling my hair trying to find a clean solution. I may just have to deal with this hacky hidden widget for now.

ZLLentz: To be clear, I'm getting a single update to the EGU value- so the alternative supported method seems to be:
- Create an EGU channel
- Connect the value_slot to a function that updates the label you care to update

But this is pretty minor I think

aberges-SLAC: I'm taking the time to refactor into over riding `_set_macro` anyway, so I think I'm going to use this as well. If I can clean up that messy block for the `update_pvs` method, it's worth it. I don't want to merge and then forget and not give this widget TLC

pcdswidgets/common/tools/feature_finder.py
KaushikMalapati: Have you tried hooking into self.motor_egu_get._channels[0].connectionStateChanged?

aberges-SLAC: I have not! But I do want to avoid accidentally propagating changes when the connection blips in and out (i.e. for las-console or during IOC reboot)

KaushikMalapati: Fair enough, we can keep timer then.

pcdswidgets/common/tools/feature_finder.py
```diff
@@ -0,0 +1,886 @@
...
+            for sig in motor_sigs:
+                widget = getattr(self, sig)
+                old_chan = widget.channel
+                new_chan = old_chan.replace(self._motor_pv, new_motor_pv)
+                widget.set_channel(new_chan)
+
+            # Properly disconnect this PV
+            self._dmov.disconnect(destroying=True)
+            # Then add the new connection
+            self._dmov = PyDMChannel(address=f"ca://{self._motor_pv}.DMOV", value_slot=self.on_motor_done_moving)
```
aberges-SLAC: Oh hey, this is a bug. It should be `new_motor_pv` not `_motor_pv`

aberges-SLAC:
Okay, I went forward with this refactor. I haven't had a chance to test my changes with real devices, however, because my test-station in the laser lab will be disconnected until next Friday

pcdswidgets/common/tools/feature_finder.py
```diff
@@ -130,6 +135,45 @@ def __init__(self, parent=None, macros=None):
...
+
+            self.connect_motor()
+
+        if macro_name == "detector":
+            self.connect_detector()
+
+        # Start the post-init timer
         self._macros_timer.start()
 
     def post_init_setup(self) -> None:
```
KaushikMalapati, line 151: I think this can be removed now?

aberges-SLAC: Actually, I use it to beat the race condition. Or at least, it should. Since `_set_macro` gets called twice during init, I want don't want to load settings twice. This slight delay should give me time to complete both sets and make sure the channels exist before it goes forward.

I'll have to make sure this actually fixes the bug I'm thinking of when I swap back and forth between macros

KaushikMalapati: I was thinking you could check if the detector and motor are connected in your overridden set_macro and if so call load settings and setup_plot there

aberges-SLAC: Since `_set_macro` has to call twice after `init`, and potentially twice during PV change, I didn't want to report the connection flag early and accidentally run this twice. I think regardless, I'm running into this issue.

ZLLentz: You can do the get_macro checks in set_macro instead of after the timer though, right? And then do the setup right away if it is needed? Something like

```
if macro_name in ["vertical", "horizontal"] and self._get_macro("vertical") and self._get_macro("horizontal"):
    ... # Do setup
```

pcdswidgets/common/tools/feature_finder.py
ZLLentz: Might be alleviated by #111

pcdswidgets/common/tools/feature_finder.py
aberges-SLAC: I think I'm trying to have my cake and eat it too — this check covers the initial setting case, but I think I _always_ run the risk of (or just straight up do) calling `load_settings` on an incomplete load. 

Gorsh, should I do something silly and add another flag attr for when the PV is changed where I do something like:
```somewhere in __init__
self._expected_macro_updates = 2
self._macro_updates = 0
```
```
if confirm == QMessageBox.Yes:
            # User confirmed, apply changes
            self._expected_macro_updates = (new_motor_pv != self._motor_pv) + (new_detector_pv != self._detector_pv)
            try:
                self.save_settings()
                self._reset_graph()
                self._set_macro("motor", new_motor_pv)
                self._set_macro("detector", new_detector_pv)
                dialog.accept()
            ...
```

Then at the end of `_set_macro()`:
```
self._macro_updates += self._macro_updates
if self._macro_updates = self._expected_macro_updates:
    self._macro_updates = 0
    self._expected_macro_updates = 0
    self.load_settings()
    self.setup_plot()
```

Is this preferred?

KaushikMalapati: Why don't we just combine the detector and motor updates into a single function and call load_settings  at the end?

aberges-SLAC: Because then I am having the same code smell where all the setup boilerplate that gets repeated from the init macro setting is unutilized when the user is changing macros, where I need to wait on `_set_macro` to be done to proceed with set-up. This becomes a "wait on the timer" thing again because `_set_macro` happens sometime after object generation and post-`__init__`.

I just mused to Zach about this and I'm not happy with my solution, but every way I can think of is equally smelly.

KaushikMalapati: Are you saying that to avoid an incomplete load, you have to wait until some set of PyDMChannels and widgets are connected before doing anything? We could

- get rid of post_init
- for all the new channels and widgets setup by _set_macro, add connection callbacks to a new method
- have the new method check if everything is connected and contains the current motor/detector pvs
- If not, do nothing, but if so call load_settings+setup_plot now that we've confirmed everything is updated and connected

but maybe that's overkill and if the current timer solution is working that's good enough.

aberges-SLAC: Yes, the incomplete load will be a bit annoying in prod because it'll end up bloating out the `settings.json` with intermediary combinations of `motor` and `detector` PVs if the user is changing both at the same time. At best this only bloats the file, at worst this accidentally overwrites valid settings for _intended_ combinations of `motor` and `detector` that we didn't want to bork on the way to setting up the new pair.

It's a pretty annoying problem.

Using `connected` callbacks is not a bad idea, I think, but I wonder if that overhead gets kinda expensive. Probably only gets `O(N)` expensive if I list-comprehension the `signal.connected` check.

I think that has less code-smell, but it adds a landmine for if the IOC reboots during the scan (for some troubleshooting reason), and then ends up re-loading old settings over the scan. Maybe I'm being too cautious?

KaushikMalapati: I think we can avoid any re-connection issues by only reloading settings if the pvs actually change. Would #110 make anything here easier?

aberges-SLAC: #110 is nice for adhering to bespoke functions, but I don't think it alleviates this particular pain point.

Also, beyond the pyqt `isSignalConnected` signal (which I'm pretty sure is _not_ the CA), there _isn't_ a nice way to hook into the `connected_changed` callback of a PyDM widget without _explicitly_ overwriting it. At that point, that's a lot of messy exceptions to handle which I'm not totally happy about.

I might just settle on this timer based code smell and accept it for now

ZLLentz: It's important to remember that, given our limited work time and allocatable scope, a PR doesn't need to be "perfect" (for some definition of perfect) to be well-designed and useful for the purposes of a merge. We can _always_ find a better way to do something, no matter how much time we put in.

I'll review once more to make sure there aren't any blocking change request sort of items.

ZLLentz (dismissed):
Possible simplifications to the setup aside, I don't see any blocking issues if you'd like to move on, Adam.

The size of this main file is getting pretty large- if we end up making other similar widgets we may choose to refactor and bring some shared code into a shared submodule.

aberges-SLAC:
Okay, the timer approach wards against intermediate loading of settings. I verified with `camonitor` that I wasn't getting any ghost channel writing as well.

https://github.com/user-attachments/assets/734ec1cd-6167-4eb3-b1ea-de8f8ecc8f77

I did have some weird timeout issue with this `LM2K4:HRM_MP0_DLY1` channel that I then couldn't immediately replicate? Nevertheless, refreshing the screen with ctrl+R would've fixed it anyhow

ZLLentz (approved):

PR #106, opened 2026-06-01
ROI selector for the Cam Viewer Widget

ipotanin:
## Description

Added ROI selection widget for the Cam Widget
- includes visibility, color, and line thickness control for the overlay
- connects to cam viewer PVS for ROI selection. 
- PV names for x,y,w,h can be macro substituted but have logical defaults
- ROI can be drawn on screen by either 2 points, changing center, or by drag and dropping 

Added a small fix the the intensity widget.
- added a block to new camera frames when user is actively dragging the brightness control

## Motivation and Context

The ASC laser lab need to often define ROI for alignment.  This is changed often and users need a smooth/friendly interface for making this selection.

## How Has This Been Tested?

Testes on two gigE cameras in the ASC.  Confirmed two-way epics->widget sync was working.

here
## Screenshots

ROI widget demo
image

## Pre-merge Checklist


pcdswidgets/imaging/common/cam_roi.py
```diff
@@ -0,0 +1,187 @@
+"""Managed rectangular ROI for camera image overlays."""
+
+from __future__ import annotations
```
ZLLentz: Question: is there a specific feature or behavior you're trying to pull in here that isn't part of base python 3.12? I lose track of which versions different things get added, but I mostly remember using this myself in python 3.9.

pcdswidgets/imaging/common/colormap_intesity_control_full.py
```diff
@@ -57,31 +59,63 @@ def __init__(self, parent: QtWidgets.QWidget | None = None):
...
+        self._on_colormap_changed(idx)
 
-        self.normalize_check.setChecked(image_view._normalize_data)
+        # Pause incoming frames while left mouse is held in histogram graph area.
+        self._histogram.viewport().installEventFilter(self)
 
         # When histogram levels change, update the image view's colormap limits
         self._histogram.sigLevelChangeFinished.connect(self._on_levels_changed)
 
+    def _pause_image_stream(self) -> None:
```
ZLLentz, line 87: Comment: I meant to look at/report this issue- seems like you were way ahead of me. I saw it when I was testing the new envs on Friday.

pcdswidgets/imaging/common/epics_roi_full.py
```diff
@@ -0,0 +1,275 @@
...
+        """Populate unset macros with sensible defaults for ROI1."""
+        default_map = {
+            "roi_plugin": ":ROI1:",
+            "suffix_X": "MinX",
+            "suffix_Y": "MinY",
+            "suffix_WidthX": "SizeX",
+            "suffix_WidthY": "SizeY",
+        }
+        for name, value in default_map.items():
+            if (name not in self._macro_values) or (self._macro_values[name] == ""):
```
ZLLentz, line 83: Question: is this ever actually called or planned to be called when the user has already set a macro? It seems to me like it's always called in `__init__`, prior to macros. Maybe it's good to be paranoid anyway.

ZLLentz (commented):
Saw some small things just to ask questions about

I think I want to play around with these a bit and include that in my review

pcdswidgets/imaging/common/cam_roi.py
ipotanin: Just to make sure I'm answering the correct question: 
    you are asking why am I overriding pg.ROI?

if yes, primarily because of recalculating bounding boxes after changing the line thickness to avoid ghosting when dragging.

pcdswidgets/imaging/common/epics_roi_full.py
ipotanin: This is how I am setting default macro values for "uncommon" settings. This way the user will see the default values in designer when they go to change it. 

Is there a better way of doing this?

pcdswidgets/imaging/common/colormap_intesity_control_full.py
ZLLentz, line 1: Side-comment: I've seen the `make` errors for the base class of this widget and I've noted them as something to address as a todo for me when I return to pcdswidgets as part of https://jira.slac.stanford.edu/browse/ECS-10276

pcdswidgets/imaging/common/cam_roi.py
ZLLentz: No, I'm asking about exactly this line here: `from __future__ import annotations`. What feature are you pulling in here?

I personally haven't needed to use it in a while so I was curious if there was something new I didn't notice.

pcdswidgets/imaging/common/epics_roi_full.py
ZLLentz: My question isn't "why do this?", it's "why be so careful?" since properties (which become macros here) are always applied after `__init__` and would overwrite these every time.

pcdswidgets/imaging/common/colormap_intesity_control_full.py
ZLLentz: (Any time somebody makes, it will undo the change you added to that file, making the specific effort applied in the last PR irrelevant)
```
zlentz@psbuild-rocky9-01:~/review/pcdswidgets(camwidgets_2 +)$ git status
On branch camwidgets_2
Your branch is up to date with 'ipotanin/camwidgets_2'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   pcdswidgets/generated/imaging/common/colormap_intesity_control_full_base.py

no changes added to commit (use "git add" and/or "git commit -a")
zlentz@psbuild-rocky9-01:~/review/pcdswidgets(camwidgets_2 +)$ git diff
diff --git a/pcdswidgets/generated/imaging/common/colormap_intesity_control_full_base.py b/pcdswidgets/generated/imaging/common/colormap_intesity_control_full_base.py
index fe9c58c..3a16bab 100644
--- a/pcdswidgets/generated/imaging/common/colormap_intesity_control_full_base.py
+++ b/pcdswidgets/generated/imaging/common/colormap_intesity_control_full_base.py
@@ -22,7 +22,7 @@ from pcdswidgets.builder.designer_widget import DesignerWidget
 from .colormap_intesity_control_full_form import *

 try:
-    from qtpy.QtCore import pyqtProperty  # noqa: F401
+    from qtpy.QtCore import pyqtProperty
 except ImportError:
     pass  # type: ignore
```

ipotanin: Ah, I see. I also ran into a CI issue for the same reason 
https://github.com/pcdshub/pcdswidgets/blob/da5722796c583ddf0f38ea9334e30e7c47264c80/pcdswidgets/generated/imaging/common/colormap_intesity_control_full_base.py#L25

pcdswidgets/imaging/common/epics_roi_full.py
ZLLentz, line 1: Suggestion: the icon buttons are nice, but they would benefit a lot from tooltips to contextualize them for first-time users.

pcdswidgets/imaging/common/cam_roi.py
ipotanin: oh, i see
I was blissfully unaware that deferred type-hint related functionality became the default behavior in python.
I will remove this line

pcdswidgets/imaging/common/epics_roi_full.py
ipotanin: very good idea! will add with other fixes.

pcdswidgets/imaging/common/camera_viewer_stretch.py
ZLLentz, line 1: Unrelated to this PR: it'd be nice to be able to have this widget be able to look at IMAGE2 or IMAGE3 instead of being hard-coded to IMAGE1

ipotanin: its part of the next pr I'm working on.  would like some feedback now since you brought it up.
I'm flip-flopping having these macro defined (easier to do, but limited functionality) , or a runtime stream configurator like this:
image

ZLLentz (dismissed):
OK, I got mega unlucky testing this PR and found a failure mode right away on default settings: if the IMAGE1 stream is pulling from ROI1 for its display, and you try to use these widgets to set ROI1, that then updates the IMAGE1 stream and most of the design assumptions here break down.

I'm not sure if this is actionable in any way directly- I think the widgets work great if the ROI doesn't change your view (though I couldn't figure out what the pen does, it seems to have no effect for me).

The only thing that needs a small amount of consideration is ROI box sizing for binned images. The detector I was using has a 4x4 binning on IMAGE1 and a full image size of 2048x2048, so when I set an roi of my visible image size (512x512) that's actually only 1/4 of the image and if I applied that ROI plugin to a different stream I'd get 1/4 of my image from a full ROI.

I think these are things to ponder about rather than something to hold back the PR.

Also, some of the resizing where the drawn ROIs catch up when you set them by number seem sluggish- not sure if this is just qt being qt.

But I really do wonder what the pen is supposed to do.

pcdswidgets/imaging/common/cam_roi.py
ZLLentz: Gotcha, I wasn't sure if there was something else I didn't know about

pcdswidgets/imaging/common/colormap_intesity_control_full.py
ZLLentz: I think what we end up doing here is omitting this entire try/except block when there isn't a pyqtProperty, e.g. when we have empty lists for what would otherwise fill up the class below

pcdswidgets/imaging/common/camera_viewer_stretch.py
ZLLentz: I'd maybe stick to the macro for the simple reason that someone is going to ask for (has already asked for) a _full camera switchout_ feature at some point (a la classic camviewer) and we could discuss it in depth and handle it at a higher level when we get there.

ZLLentz: One more thought to contradict myself: would it be useful to set an ROI here using the full-frame image and then switch the view to the smaller image later? Maybe

ipotanin:
Definitely will include the hover hints for the buttons in this PR, as the pen use was not as obvious as I thought 🙃 (it's a two click draw box, firstclick-> top left, second click-> bottom right)

Would love to brainstorm how to handle streams being fed by ROIs. My use case for this is that ROI is not affecting the stream, but clearly could use more thought.

Box binning was very much on my mind and was a "can that I kicked down the road". would love to brainstorm this too!

pcdswidgets/imaging/common/epics_roi_full.py
ipotanin: Might be worth a quick offline chat on this. I think I may not have the correct understanding on how (the order of which) macros resolve.

pcdswidgets/imaging/common/colormap_intesity_control_full.py
ZLLentz: I'm gonna blitz out a PR for this I think it should be simple

ZLLentz:
RE: the pen, I think what tripped me up is:
- I'm used to pen-like tools in other applications being click/drag
- It's hard for me to keep the mouse still and not drag
- Is there a visual indicator that the first click did something?

In addition to accidentally stumbling upon a bad test example

ZLLentz (approved):

ZLLentz:
RE: ROI affecting the stream: I think you'd normally be building screens with the area detector config in mind, so as long as the documentation is clean this can be avoided.

In my case, I picked a random camera that wasn't set up for this widget (to test it) and got burned a bit from a non-standard configuration

ipotanin:
I was on the fence about click-drag vs two-click. 

am down to change it to click-drag in a future PR (should not be much effort).  I'll merge as-is for forward progress

ipotanin:
RE: binning and ROIs affecting the stream.

I wondered if we have a documented "standard" way we like to configure streams (and if we could potentially change this for future projects)?

My thoughts on what defaults should be (just brainstorming):
```
DATA1:  always full CAM. never touch its stream specific ROI and binning. and possibly remove these sub-plugins to enforce this. 
IMAGE1 :  always full CAM input, adjust binning here in the sub plugin, but do not adjust roi on this stream

extra plugins always set up
ROI1-4: CAM input, only adjust software ROIs here
STATS1-4: ROI1-4 input

add additional streams to view these ROI
ROI_VIEW1-4:   input is ROI1-4

```
I was thinking of enforcing this with the next gigECam common, going as far as to disable any stream or plugin that is not this unless explicit in the cfg file.

No action items here. just wanted to put my thoughts in words.

ZLLentz:
We don't have a documented standard and we're well overdue for having one.

PR #107, opened 2026-06-02
MNT: only include pyqtProperty import if it is used to avoid lint errors

ZLLentz:
## Description

In `ui_base_widget.j2`, we had an import that was always included, even when it wasn't used, causing some linting/CI issues. A simple resolution is to no longer include the import when it is not needed.

Therefore, I surrounded the import in an `if` block and matched it to the same variable that is iterated through below when the import is used.

## Motivation and Context

Errors/warnings on `make`, CI issues, general annoyance

## How Has This Been Tested?

Interactively plus the unit tests pass locally

Here only

## Pre-merge Checklist


KaushikMalapati (approved):

ipotanin (approved):
LGTM

PR #108, opened 2026-06-03
MNT: use triple quotes to allow user to include single quotes in input

ZLLentz:
## Description

- Whenever we take a user value and store it as a runtime-editable macro, wrap this value in triple quotes.
- This allows the user to pick values that contain single quotes in them.
- Apply these features to tip/tilt to slightly simplify the setup

Also:
- Add an option in the Makefile to do a full rebuild of the `generated` filetree, so I could commit those here instead of making someone else do it.

## Motivation and Context

From a conversation with @aberges-SLAC where he had to find a workaround because one of his input values for a macro needed to contain a quotation mark.

This ended up as a line-item in https://jira.slac.stanford.edu/browse/ECS-10276 which I'm trying to go through now

## How Has This Been Tested?

The test suite ingests this a bit and it seems OK
I created a test screen and it seems to work OK

Here only, for now

## Screenshots

image

image

## Pre-merge Checklist


KaushikMalapati (dismissed):

aberges-SLAC:
I remember this conversation, but for the life of me I cannot remember how I found this edge case

aberges-SLAC:
AH no, this was actually for the `expert_button` args which needed to be passed a json dict of `"{motor}": "${vertical_motor}"` and I just ended up doing this manually an ugly way in the widget

pcdswidgets/builder/ui_base_widget.j2
```diff
@@ -55,14 +55,14 @@ class {{ base_cls }}(DesignerWidget):
 {% for widget in widget_names %}
         "{{ widget }}": [
 {% for prop, value in widget_to_pre_templ_strs[widget] %}
-            ("{{ prop }}", "{{ value }}"),
+            ("{{ prop }}", """{{ value }}"""),
```
aberges-SLAC, line 58: Pretty sure this was the culprit when I ran into:
```
invalid-syntax: Expected `,`, found `{`
   --> pcdswidgets/generated/motion/smaract/smaract_tip_tilt_double_base.py:115:19
    |
113 |                 "macros",
114 |                 [
115 |                 ""{'motor'": "'${horizontal_motor}'}"",
    |                   ^
116 |                 ]
117 |             ),
    |
```

So I approve! I have messages when we checked this out and you had me edit this line directly

aberges-SLAC (dismissed):

aberges-SLAC:
> AH no, this was actually for the `expert_button` args which needed to be passed a json dict of `"{motor}": "${vertical_motor}"` and I just ended up doing this manually an ugly way in the widget

This issue, however, might still need to be handled by:
```
self.vertical_expert_screen.setMacros(['{"motor": ' f'"{self._get_macro('vertical_motor')}"' "}"])
        self.horizontal_expert_screen.setMacros(['{"motor": ' f'"{self._get_macro('horizontal_motor')}"' "}"])
```

ZLLentz:
I was hoping we could now just have something like this in the ui file:
```
{"motor": "${horizontal_motor}"}
```
I should play around with it a bit and see if that's the case or not before merging

ZLLentz:
Yeah, it works now with just the ui file if you put in valid json text- I applied this to tip-tilt here and removed the extra macro sets to show that it works

ZLLentz:
Maybe the combination of this and #109 makes the setup for these two tip/tilt widgets a lot simpler- even if others still need some extra, my goal is for "normal" pydm features to also work here.

KaushikMalapati (approved):

PR #109, opened 2026-06-03
ENH: allow relative filepaths in widget ui files

ZLLentz:
## Description

At runtime, check for `PyDMEmbeddedWidget` and `PyDMRelatedDisplayWidget` and update any `pcdswidgets/` paths included to be absolute paths relative to the pcdswidgets repo root.

In addition:
- Update `SmaractTipTiltDouble` and `SmaractTipTiltFull` to use this as an example and interactive test.
- Fix a bug related to string lists with empty elements that was caught by one of the new tests

## Motivation and Context

It's not possible to express "refer to this file in pcdswidgets" right now. Instead of resolving them relative to this repo's install location, they are resolved relative to the location of the ui file that contains the widget when used. The changes here make it possible to define these as paths relative to the pcdswidgets install without resolving them manually at runtime in every widget that needs this behavior.

@aberges-SLAC had worked around this while adding expert screen buttons to his smaract tip/tilt UIs. The diff in those files in the PR here undoes his workarounds that are no longer needed.

## How Has This Been Tested?

Interactively with a screen (see screenshots below).
I extended the test widgets to have tests for both this feature and for the quote escaping.

Here only

## Screenshots

image

image

## Pre-merge Checklist


ZLLentz:
@aberges-SLAC is it worth me pursuing this to a merge? This is a proof of concept to address one of the issues you had while setting up the tip/tilt widgets

aberges-SLAC:
I think this is definitely worth pursuing, it's a clever patch!

I haven't explored the widgets Ilia added too deeply, so I'm unaware of the impact there. However, my hope is that this simplifies it significantly.

It also reminds me to come back to these widgets when I have a real ioc-level `expert screen` that can be spawned, since there are actually quite a lot of signals someone will want for a detailed SmarAct screen

ZLLentz:
I'm going to add a unit test here and likely will need to triage merge conflict stuff with #108

ZLLentz:
I'm also going to cover #108's feature in this unit test- these are a bit easier to test together than separate.

pcdswidgets/builder/designer_widget.py
```diff
@@ -100,6 +105,42 @@ def get_designer_icon(cls) -> QIcon | None:  # type: ignore
...
+        /some/path/python/lib/site-packages/pcdswidgets
+        and we pick a relative path motion/smaract/test.ui
+        then we will replace it with the absolute path
+        /some/path/python/lib/site-packages/pcdswidgets/motion/smaract/test.ui
+        """
+        # ui-file loaded objects are attributes of the class
+        for obj in self.__dict__.values():
+            if isinstance(obj, PyDMEmbeddedDisplay):
+                fname = obj.readFilename()
+                if os.path.isabs(fname):
```
KaushikMalapati: I think this will return false is the filename is unset and we'll end up with just MODULE_ROOT as the path, althought in practice I think it will always be defined

KaushikMalapati:
Are there any cases where one would want a relative path that is not relative to pcdswidgets, like I have another with ui files one of which has a PyDMRelatedDisplayPath that references another ui file in the same repo relatively?

pcdswidgets/builder/designer_widget.py
```diff
@@ -72,6 +76,7 @@ def __init_subclass__(cls):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.ui_form.setupUi(self, self)  # type: ignore
```
KaushikMalapati, line 78: What does this do? I wonder if it makes sense to update paths before setup

aberges-SLAC:
> Are there any cases where one would want a relative path that is not relative to pcdswidgets, like I have another with ui files one of which has a PyDMRelatedDisplayPath that references another ui file in the same repo relatively?

Yes, what if I use a `pcdswidget` to make a larger context screen/app/widget where I want to override the expert screen button to use the relative path in my repo?

e.g. launchgui screens in a common IOC

aberges-SLAC (commented):
FWIW I am a big fan of the simplification this brings to the tip-tilt widgets.

pcdswidgets/builder/designer_widget.py
aberges-SLAC: It's how the generated widget bases spawn the UI objects, e.g.:
https://github.com/ZLLentz/pcdswidgets/blob/d869156ea72e4b31e2757823d02b44cf3a79eda8/pcdswidgets/generated/motion/smaract/smaract_tip_tilt_full_form.py#L19-L462``

pcdswidgets/tests/builder/test_designer_widget.py
```diff
@@ -115,6 +139,27 @@ def test_macro_substitution_list_widget(test_widget: WidgetForBuilderTest):
...
+    test_widget.setProperty("rel_title", "Related")
+
+    assert test_widget.rel_disp.readMacros() == ['{"TITLE": "Related"}']
+
+
+def test_filepath_subdisplays(test_widget: WidgetForBuilderTest):
+    canonical_path = Path(__file__).parent.resolve() / "subdisplay.ui"
+    assert canonical_path.exists()
+    assert test_widget.emb_disp.readFilename() == str(canonical_path)
+    assert test_widget.rel_disp.readFilenames() == [str(canonical_path)]
```
KaushikMalapati: Is it worth adding an absolute path to rel_disp in widget_for_builder_test.ui to "test" that it doesn't get prepended with anything?

pcdswidgets/tests/builder/test_designer_widget.py
KaushikMalapati, line 1: There is a lot of boilerplate you had to add here just to put some new widgets into widget_for_builder_test.ui

pcdswidgets/builder/designer_widget.py
ZLLentz: I was thinking about this but didn't decide what I'd want to do. The widgets never work as expected with an empty filename in any contexts, so I wasn't sure how much effort I should put into checking it here.

pcdswidgets/tests/builder/test_designer_widget.py
ZLLentz: Yes it is worth it, I will add it as the second path

pcdswidgets/tests/builder/test_designer_widget.py
ZLLentz: I think in principle I could have skipped most of it and probably will in the future

pcdswidgets/builder/designer_widget.py
KaushikMalapati: Fair enought, I think it might only ever be blank if the files are set based on user interaction

ZLLentz:
> > Are there any cases where one would want a relative path that is not relative to pcdswidgets, like I have another with ui files one of which has a PyDMRelatedDisplayPath that references another ui file in the same repo relatively?

> Yes, what if I use a pcdswidget to make a larger context screen/app/widget where I want to override the expert screen button to use the relative path in my repo?

> e.g. launchgui screens in a common IOC

Should I make this _require_ the `pcdswidgets` prefix in the path to allow for this?

`ui/thing/stuff` -> relative to my screen
`pcdswidgets/ui/thing/stuff` -> relative to the pcdswidgets install

I think that's sufficiently unambiguous and allows for less unexpected behavior with paths. It also would let you leverage `PYDM_DISPLAYS_PATH` if you're into that sort of thing.

pcdswidgets/builder/designer_widget.py
ZLLentz: Right, this is where the widgets that I'm manipulating actually appear for the first time.

pcdswidgets/tests/builder/test_designer_widget.py
ZLLentz: I might even undo some of it to make the tests more readable again

pcdswidgets/builder/designer_widget.py
KaushikMalapati: So won't setupUI cause them to be setup with invalid paths before update_relative_paths is called and makes the affected widgets re-render?

aberges-SLAC:
> > > Are there any cases where one would want a relative path that is not relative to pcdswidgets, like I have another with ui files one of which has a PyDMRelatedDisplayPath that references another ui file in the same repo relatively?
> 
> > Yes, what if I use a pcdswidget to make a larger context screen/app/widget where I want to override the expert screen button to use the relative path in my repo?
> 
> > e.g. launchgui screens in a common IOC
> 
> Should I make this _require_ the `pcdswidgets` prefix in the path to allow for this?
> 
> `ui/thing/stuff` -> relative to my screen `pcdswidgets/ui/thing/stuff` -> relative to the pcdswidgets install
> 
> I think that's sufficiently unambiguous and allows for less unexpected behavior with paths. It also would let you leverage `PYDM_DISPLAYS_PATH` if you're into that sort of thing.

I like this solution: readable, as succinct as it gets, and gives us enough rope to hang ourselves with

pcdswidgets/builder/designer_widget.py
ZLLentz: This definitely affects the embedded widget in some way- I'll try to understand what the behavior is. I think the related display button is safe because it doesn't check the file until you click it.

Prior to `setupUi` there isn't any object to manipulate though, so I'm not exactly sure what to do. There might be some magic with the file generators we can do.

ZLLentz:
I'm imagining all the ways that people accidentally trip into this behavior and I think requiring `pcdswidgets/` must be the way to go (unless there's an even more magical invocation to use)

KaushikMalapati:
> Should I make this _require_ the `pcdswidgets` prefix in the path to allow for this?

Yes. Anyone who has relative paths starting with pcdswidgets but thinks they shouldn't point to the library they're using with the same name is asking for trouble.

ZLLentz:
Ok, that's the path I'll follow tomorrow. I'm giving up on being focused today.

ZLLentz:
Thank you for the helpful reviews

pcdswidgets/tests/builder/test_designer_widget.py
ZLLentz: I'll do what I should have done from the start: leave the original test widget alone, and make some new test widgets for the new tests. This keeps them more separate and should be a more sustainable testing model.

pcdswidgets/builder/designer_widget.py
ZLLentz: It seems like the main consequence here is that some time is wasted trying to load the nonexistent file (or, rarely, the wrong file) before we switch the filename for the embedded display.

The only ways I know to work around this are:
- Create a custom embedded display widget that people can use that implements the special behavior
- monkeypatch the embedded display class during form loads
- parse and edit the form.py files during the build steps

If we think these are valuable I'll suggest that we do it another time.

ZLLentz:
I think I'm happy with this now, and incidentally found another small bug to patch while doing it

pcdswidgets/builder/designer_widget.py
ZLLentz: This ends up being no longer relevant with the new implementation

pcdswidgets/builder/build.py
```diff
@@ -212,7 +212,12 @@ def add_prop_to_widget_macros(widget_macros: defaultdict[str, dict[str, str | li
     if strlist_node is not None:
         # We have a list of strings! Some may have macros.
         all_str_nodes = strlist_node.findall("string")
-        all_str_literals = [node.text for node in all_str_nodes if node.text is not None]
+        all_str_literals = []
+        for node in all_str_nodes:
+            if node.text is None:
+                all_str_literals.append("")
+            else:
+                all_str_literals.append(node.text)
```
KaushikMalapati, line 220: Why is this necessary? ${ will never be in any ""

pcdswidgets/builder/designer_widget.py
KaushikMalapati: It's not that valuable, I forgot we needed setupUI before we could easily access the widgets.

KaushikMalapati (approved):

pcdswidgets/builder/build.py
ZLLentz: This is a bugfix for the following situation:

- string list
- one element of the string list has a macro
- other elements don't, including at least one empty string

What was happening before is that the empty strings would be dropped from the result:

`["${macro}", ""] -> ["${macro}"]`

Which caused a problem for the related display test widget where I didn't have macros to include for the relative and absolute path examples (to the fake screens), so the macros list dropped all the empty components.

PR #110, opened 2026-06-04
ENH/MNT: macro helpers

ZLLentz:
## Description

- Promote the "main" macro getter/setters to public methods to clarify that this is the desired way to get/set them
- Add an `after_set_macro` overridable function for setting up additional widget resources when a macro is set
- Implement `after_set_macro` on the tip/tilts as an example of how I imagine these would be used

## Motivation and Context

Another line item from https://jira.slac.stanford.edu/browse/ECS-10276
based on a conversation I had with @aberges-SLAC 

## How Has This Been Tested?

Interactively only, things seem about the same
The tests still pass

Here only

## Pre-merge Checklist
- [ ] ~New/changed widgets are part of the test suite (semi-automatic)~


ZLLentz:
One more proof of concept for your consideration @aberges-SLAC - worth pursuing to the end, or not useful for what you were trying to do?

aberges-SLAC:
> One more proof of concept for your consideration @aberges-SLAC - worth pursuing to the end, or not useful for what you were trying to do?

Exactly useful for what I was trying to do and stops me from hurting myself in my confusion when overriding set_macro

ZLLentz:
TODO:
- Merge recent updates
- Resolve conflicts
- Make again
- Add back _get_macro and _set_macro with deprecation warning so we don't step on our own toes too hard here

pcdswidgets/builder/designer_widget.py
```diff
@@ -100,18 +101,45 @@ def get_designer_icon(cls) -> QIcon | None:  # type: ignore
             case _:
                 return None
 
-    def _get_macro(self, macro_name: str) -> str:
+    def get_macro(self, macro_name: str) -> str:
+        """Returns the current value of a macro that is applied to our subwidgets."""
         return self._macro_values[macro_name]
 
-    def _set_macro(self, macro_name: str, value: str):
+    def _get_macro(self, macro_name: str) -> str:
```
KaushikMalapati, line 155: Do we really need the deprecated versions? I think _set_macro is not used anywhere and you already remove the only _get_macro references in the two smaract classes

pcdswidgets/builder/designer_widget.py
```diff
@@ -100,18 +101,45 @@ def get_designer_icon(cls) -> QIcon | None:  # type: ignore
...
+
+    def _set_macro(self, macro_name: str, value: str):
+        warnings.warn("_set_macro is deprecated, use set_macro instead.", category=DeprecationWarning, stacklevel=2)
+        return self.set_macro(macro_name, value)
+
+    def after_set_macro(self, macro_name: str, value: str):
+        """
+        Hook for additional actions to take when a macro is set.
+
+        This exists so you can extend this behavior without overriding set_macro,
```
KaushikMalapati, line 173: How could overriding _set_macro break the widget? I would think that people might want to do it anyways in case they ever need to do something before _macro_values is updated

pcdswidgets/builder/designer_widget.py
ZLLentz: It's only for other open/lurking PRs to give a gentle landing rather than a hard break. I could go either way.

pcdswidgets/builder/designer_widget.py
ZLLentz: I broadly agree that nothing in this PR is necessary- stuff like this hook only exist because it feels nicer to have someone say "this thing you're doing, it's very normal, this is where you should implement your post-macro-setting behavior", rather than the `_set_macro` override which feels like circumventing a private method or a "solid" foundation that ought not to be disturbed.

In a very literal sense this is frivolous, but I know sometimes we'd forget to e.g. call `super()` and get confused.

KaushikMalapati: Fair

KaushikMalapati (approved):

PR #112, opened 2026-06-05
BLD/DOC: switch to and require pixi, remove extra boilerplate, update docs

ZLLentz:
## Description

We've been using a stacked environment: `uv` on top of a temporary base env that I threw together some months ago.
This switches out the stacked `uv` environment for an entirely self-contained `pixi` environment.

This also removes the option to do development work in this repo without the `ctrlenv` tools on your path - you'll at least need `pixi`.

I considered removing the `try_in...` scripts in favor of just `pixi run pydm`, `pixi run designer`, but I thought it might be nice to keep them.

This update is only possible after the learning from the pixi environments elsewhere.

## Motivation and Context

- The previous setup required a reference to the temporary base env, which I will certainly accidentally delete one day.
- We could have simply moved the stack to a new base, but I think for dev environments keeping them fully separate gives us a lot more flexibility.
- This one is self-contained: after install, the environment will continue to work "forever".
- We need the `conda-forge` builds of `pyqt` and `python` to keep the `designer` plugin working- maybe one day we'll solve or otherwise work around this problem, but that's not today.
- `uv` wasn't enough on its own to solve the packaging here (for dev work) without a base env- it doesn't let you install qt itself, for example
- This incidentally solves a line item issue for `pcdswidgets` where if you have `pcds_conda` active all of this would stop working. That's no longer the case because the `uv` build in `pcds_conda` won't be pulled in here again.

## How Has This Been Tested?

Interactively everything still works as I expect

In the README

## Pre-merge Checklist
- [ ] ~New/changed widgets are part of the test suite (semi-automatic)~


ZLLentz:
I do want some opinions here from people who have been using the uv variant of this repo

ipotanin (dismissed):
LGTM

I cloned and tried it myself and it seemed to work. 

On the note of the try_in scripts, what does the --as-is pixi flag do?  

I think it's fine to keep the try in scrips, but I do see how now they don't do much and it may be better to get people in the habit of pixi run instead!

side note, I ran out of space in my home dir when trying to build and had to del some old files.  this repo with the env built is 1.3Gb.

ZLLentz:
> On the note of the try_in scripts, what does the --as-is pixi flag do?

`--as-is` directs `pixi` to not check for possible needed updates to the local install- I guess this isn't actually necessary here. I was copying some scripts from the `ctrlenv` stuff where I have some routines that need to be runnable when the environment is write-protected, but that's not the case for these.

> it may be better to get people in the habit of pixi run instead

I'm starting to feel the same, we can sleep on it over the weekend

>  I ran out of space in my home dir when trying to build and had to del some old files. this repo with the env built is 1.3Gb

This gives me at least a little bit of pause, I wonder if we'll have to start doing python work that includes environments in a non-home area (similar to what we do with `epics-dev`)

ZLLentz:
Returning today, I intend to make the following changes:

- Remove `--as-is` so that these scripts would work out-of-the-box after cloning
- Add a deprecation warning to the two scripts directing people to use `pixi run` to help us make a soft landing as we muscle memory our way into running the scripts, with the intention of deleting them later

I think that puts us in a reasonable spot

ZLLentz:
I'm hoping to get this in- any other thoughts or concerns?

ipotanin (approved):
LGMT

aberges-SLAC:
I too am a little spooked by the storage space issue  in `~` given the various NFS issues we've had this year. Also, the not-so-distant past where we critically needed to reallocate space to `/cds/`. 

Maybe this is less of an issue in the `/sdf/` transition?

Also, it took me ≈ 10 mins to do the initial `pixi run install` of a fresh clone of this branch. Subsequent `make`s are probably really quick

aberges-SLAC (approved):

aberges-SLAC:
For more context btw:
```
pcdshub$  du -sh *
1.3G    bld_pixi
12M     dream_smaract_slits
5.6M    engineering_tools
1.9M    lcls-pydm-common
21M     pcdsdevices
224M    pcdswidgets
11M     testing
6.4M    typhos
```

That is _pretty_ large comparatively. It's rivaling the storage space of my entire `ioc` dev directory, which spans many area specific templates and common IOCs/modules.

ZLLentz:
Yeah, I'm thinking it may need to be common practice to develop these repos in `epics-dev` or in a new `pyps-dev` area of sorts, to keep the home areas clean. This won't be unique to this repo as we switch away from central envs.

ZLLentz:
Probably we'll also want to set up environment variables for user-specific cache directories on the same disk as we do dev work on, so uv/pixi can hardlink files and save space overall (even if our home areas are liberated)

PR #113, opened 2026-06-11
Clean merge of #104 (slits widget, layout/font updates for motor widgets)

ZLLentz:
This takes the contents of #104 and cleanly merges them into master with no functional changes.

I'm going to merge this without review as if I had just clicked the merge button in #104.

PR #114, opened 2026-06-11
ENH: add property to MotorClassicFull to change error reset PV/value

ZLLentz:
## Description

Add a `motor_type` enum property to `MotorClassicFull` that determines which PV and value are used for the error reset widget.

There are currently four options:
- `GENERIC`, which simply hides the error reset button
- `IMS`, which uses the pv name and value from before this PR
- `BECKHOFF`, which uses the pv name from the new motion library and set value 1
- `BECKHOFF_LEGACY`, which uses the pv name from the old motion library and set value 1

## Motivation and Context

- Aside from the clear error button, this widget is fully generic.
- This PR makes the widget fully generic: we can use the same layout for any motor type, just switching out the PV and value to use.
- This pattern would allow us to make other changes later, e.g. customizing what we call for the expert screen if we'd like to.
- The enum boilerplate is pretty ugly but I think this is much nicer than typing strings or ints into the designer property.

This is a follow-up to #104, #113

## How Has This Been Tested?

Interactively only, I created some soft iocs and made sure the right values are being put to by the error reset widget. I don't have a reasonable way to test this in prod non-disruptively.

Here only

## Screenshots

image

image

## Pre-merge Checklist


ZLLentz:
I need to fix the test which makes some assumptions about the contents of these widget files

pcdswidgets/motion/common/motor_classic_full.py
```diff
@@ -4,14 +4,121 @@
...
+    # Boilerplate to make the enum property work
+    if ACTIVE_QT_WRAPPER == QtWrapperTypes.PYQT5:
+        from PyQt5.QtCore import Q_ENUM
+
+        Q_ENUM(MotorTypes)
+    MotorTypes = MotorTypes
+    GENERIC = MotorTypes.GENERIC
+    IMS = MotorTypes.IMS
+    BECKHOFF = MotorTypes.BECKHOFF
+    BECKHOFF_LEGACY = MotorTypes.BECKHOFF_LEGACY
```
KaushikMalapati, line 81: What are these for?

KaushikMalapati (approved):

pcdswidgets/motion/common/motor_classic_full.py
ZLLentz: It's confusing, but without these defined qt explodes when trying to run the pydm screen.
I've removed them and ran again to paste the error here and to construct a more complete explanation/understanding:

<details>

```
Traceback (most recent call last):
  File "/cds/home/z/zlentz/projects/widget_cleanup_ecs-10276/pcdswidgets/.pixi/envs/default/bin/pydm", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File "/cds/home/z/zlentz/projects/widget_cleanup_ecs-10276/pcdswidgets/.pixi/envs/default/lib/python3.12/site-packages/pydm_launcher/main.py", line 120, in main
    app = pydm.PyDMApplication(
          ^^^^^^^^^^^^^^^^^^^^^
  File "/cds/home/z/zlentz/projects/widget_cleanup_ecs-10276/pcdswidgets/.pixi/envs/default/lib/python3.12/site-packages/pydm/application.py", line 134, in __init__
    self.main_window.open(ui_file, macros, command_line_args)
  File "/cds/home/z/zlentz/projects/widget_cleanup_ecs-10276/pcdswidgets/.pixi/envs/default/lib/python3.12/site-packages/pydm/main_window.py", line 412, in open
    new_widget = load_file(filename, macros=macros, args=args, target=target)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/cds/home/z/zlentz/projects/widget_cleanup_ecs-10276/pcdswidgets/.pixi/envs/default/lib/python3.12/site-packages/pydm/display.py", line 73, in load_file
    loaded_display = loader(file, args=args, macros=macros)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/cds/home/z/zlentz/projects/widget_cleanup_ecs-10276/pcdswidgets/.pixi/envs/default/lib/python3.12/site-packages/pydm/display.py", line 211, in load_ui_file
    display.load_ui_from_file(uifile, macros)
  File "/cds/home/z/zlentz/projects/widget_cleanup_ecs-10276/pcdswidgets/.pixi/envs/default/lib/python3.12/site-packages/pydm/display.py", line 454, in load_ui_from_file
    _load_compiled_ui_into_display(code_string, class_name, self, macros)
  File "/cds/home/z/zlentz/projects/widget_cleanup_ecs-10276/pcdswidgets/.pixi/envs/default/lib/python3.12/site-packages/pydm/display.py", line 184, in _load_compiled_ui_into_display
    klass.setupUi(display, display)
  File "<string>", line 22, in setupUi
AttributeError: type object 'MotorClassicFull' has no attribute 'GENERIC'
```

</details>

So, in the moment I simply added them in to make the error go away (and because my examples in PyDM do the same), but going further this must be due to how the pyuic5 ui file compiler works. This is done at runtime for PyDM so we can't examine the source code unless we generate it ourselves:

```
pixi run pyuic5 -o ../output.py ../test_motor_err.ui
```

Scrolling to lines 20-25 as per the traceback where we error on line 22

```
        self.MotorClassicFull = MotorClassicFull(Form)
        self.MotorClassicFull.setToolTip("")
        self.MotorClassicFull.setProperty("motor_type", MotorClassicFull.GENERIC)
        self.MotorClassicFull.setObjectName("MotorClassicFull")
        self.verticalLayout.addWidget(self.MotorClassicFull)
```

It becomes clear that the pyuic5 compiler that generates the python code has a baked-in assumption that all enum members are also stored as class variables. So this must be the underlying reason why we have to include these splayed out in the class here.

PR #115, opened 2026-06-19
Vacuum Expert Screens

TrumanPar:
## Description

I have added a few helper functions to the PCDSSymbolBase class that enforces a new expert screen pattern:

**get_expert_ui_paths:**
This should be overridden in the derived class. It returns a list of strings that resolves to where the .ui files are stored in the repo.

**get_expert_macros**
Likewise, should be overridden, returns a dictionary of named macros and the resolved string representation. By default will populate prefix.

**get_expert_tab_specs**
Calls the above two functions to build a list of dictionary entries. Each entry contains the .ui path, macros, and screen title. 

**_format_macros**
Static class that formats macros returned by `get_expert_macros` into strings that pydm can work with, called in `get_expert_tab_specs`.

I also modified `_handle_icon_click` to build a tabbed widget using the above helpers. Existing logic is left intact.

**Mixins:**
The interlock mixin also needed a fix. Since I no longer use an interlock for this widget when using `PIPCombined`, I thought it was appropriate to pass False to the interlock suffix. `create_channels` would just return early before and keep the old stale interlock state, but now I force the channel to disconnect, set it to None, force the self._interlocked bool to False, and enable the control frame. This makes it so the stale interlocked suffix is overridden, the widget no longer shows the interlock state, and the controls are available. But, I worry this will introduce issues - what if the suffix becomes false for another reason, one that isn't user driven? Is this even possible? Also, the other mixins still just return early so the pattern across mixins is broken. We should evaluate if this is a safe thing to implement, I am leaning towards no but I'm not sure how else I can 'disable' a mixin cleanly.

**IonPump changes:**
Currently, this widget is setup for our PLC PVs, but we want to have the option to utilize the ion pump controller PVs as well (see more in motivation and context). I imagine this is a good pattern so we can tailor the widgets to source from just about any set of PVs and we won't need to make completely new icons, improving scalability. 

I accomplish this with the `SUFFIX_MAP` which is a dict that maps ophyd devices to the appropriate PV suffixes. The key can really be any string, but I want to use the existing `EXPERT_OPHYD_CLASS` property so this makes for a smooth transition away from typhos. I also added the override that will change the default suffixes out. If the key is not found, it defaults to the initial suffixes.

Then, I just implemented `get_expert_ui_paths` and `get_expert_macros`. I had to import caget to accomplish the macro expansion, I hope this isn't an issue. 

I also have added .ui files and I have taken the liberty of how they should be setup. I was thinking that `ui/vacuum/<device>_screens/<ophyd class name>_<screen title>` makes for a decent pattern. I had to append `_screens` to the directory name otherwise it shadows the name of pumps.py and pixi complains when it builds. 

I still need to review with Jing that all the screens match her expectations. 

## Motivation and Context

My goal for this PR is to establish a strategy for building expert screens for the symbol based vacuum widgets now that we're not using `Typhos`. I am starting with the `IonPump` class as a proof of concept.

This also aligns with a separate effort to change how we control the Gamma Ion pumps. We want to allow the QPC controller itself to interlock the pumps based on its own pressure reading and pressure setpoint. The PLC should only read in a status and a pressure. Therefore, we are introducing a second version of the PLC FB that no longer outputs HV_ON to the controller, will never interlock the pump, and will only ever show the analog reading. On and off commands will also emanate from the controller IOC now that the PLC has no outputs. This is why I needed to add the suffix map to the derived class to grab PVs from both the PLC and the IOC.

## How Has This Been Tested?

Interactively with existing ion pumps

## Screenshots

Below is an image of the widget loaded in pydm:
image

And a screenshot in designer:
image

Representative shot of the PIPCombined expert screen, all screens populate and display correctly.

image

And the legacy PIPPLC widget's new screens (only detailed shown):
image

## Pre-merge Checklist


KaushikMalapati:
@TrumanPar the names of the generated files don't match the new ui files you added. Can you try remaking locally and getting pytest to pass?

TrumanPar:
Yeah I noticed that, I tried to rebuild but they didn't get generated. Looking into why

TrumanPar:
@KaushikMalapati I restructured my ui file path to the depth specified in the makefile and it worked. I had to change the implementation of `get_expert_ui_paths` but I don't think it adds complexity. 

I just need to retest that everything is working with a running ion pump.

TrumanPar:
Retested and the screens look good.

TrumanPar:
@KaushikMalapati let me know what you think about how I go about 'disabling' the interlock mixin. Not sure if this is the correct approach.

ZLLentz:
Comments about the PR description before I dive into the code:

> 'disable' a mixin cleanly

I think the normal thing to do here would be to have separate classes with and without the mixin (haven't read through the code yet, might not apply)

> I had to append _screens to the directory name otherwise it shadows the name of pumps.py and pixi complains when it builds.

I'm starting to think we should have had separate ui directories for static screens to be loaded by PyDMRelatedDisplay, different from the current ui directory which is only for generated screens. I think it feels natural for people to put all the ui files there which is causing some problems.

Motivation makes sense, we chatted about this offline at some point

I'll read the code now

pcdswidgets/vacuum/base.py
```diff
@@ -588,6 +588,89 @@ def assemble_layout(self):  # noqa: C901
             box_layout.addWidget(widget)
             layout.addLayout(box_layout)
 
+    def get_expert_ui_paths(self, expert_key) -> list[str]:
```
ZLLentz: Nitpick: your type hints are only partially completed in many methods in the class here, missing the input types. I'm used to them either being fully complete or not attempted at all, so this surprised me.

pcdswidgets/vacuum/base.py
```diff
@@ -588,6 +588,89 @@ def assemble_layout(self):  # noqa: C901
...
+
+        Parameters
+        ----------
+        expert_key : str
+            The expertOphydClass value.
+
+        Returns
+        -------
+        list[str]
+            Absolute paths to local UI files. Return an empty list when no
```
ZLLentz: Docstring nitpick: I suspect you don't actually want the user to include an absolute path here.

Edit: I see later that you did intend this, but I think this is the wrong idea. We ought to not force every widget to duplicate the code for resolving local screen dirs.

pcdswidgets/vacuum/base.py
```diff
@@ -626,35 +709,46 @@ def _handle_icon_click(self):
...
 
-        try:
-            import typhos
-        except ImportError:
-            logger.error("Typhos not installed. Cannot create display.")
-            return
+        # Collect the tabs to show in order:
+        # 1. The local .ui file mapped to the expertOphydClass - first tab(s).
+        # 2. Any designer-provided ui_* files - following tab(s).
+        # 3. Typhos, if it is installed - last tab.
```
ZLLentz, line 741: Comment: I like this behavior :+1:

pcdswidgets/vacuum/base.py
```diff
@@ -626,35 +709,46 @@ def _handle_icon_click(self):
...
+            import typhos
 
-            self.tab_widget.show()
+            kwargs = {"name": name, "prefix": prefix}
+            display = typhos.TyphosDeviceDisplay.from_class(klass, **kwargs)
+            self._expert_display = display
+            display.destroyed.connect(self._cleanup_expert_display)
+            self.tab_widget.addTab(display, "Typhos")
+        except ImportError:
+            logger.warning("Typhos not installed. Skipping Typhos display.")
```
ZLLentz: Suggestion: demote this to debug, this will appear a lot in the terminal for some screens otherwise.

pcdswidgets/vacuum/pump_screens/PIPCombined_controller.py
ZLLentz, line 1: Question: did you intend to include all of these as standalone widgets? It seems strange to me to have these as designer widgets. Maybe this is fallout from the ui folder generating everything as if it was to be used standalone in designer.

pcdswidgets/vacuum/pumps.py
```diff
@@ -99,9 +116,106 @@ def __init__(self, parent=None, **kwargs):
...
+        # Rebuild channels so the new suffixes take effect (only meaningful
+        # once a channel prefix has been provided).
+        if self._channels_prefix:
+            self.destroy_channels()
+            self.create_channels()
+
     def sizeHint(self):
         return QSize(180, 80)
 
+    def get_expert_ui_paths(self, expert_key):
```
ZLLentz: Suggestion: we need to make the API simpler here. You've had to add lot of code for this (and we'll need to do it again for every widget) when we ought to just be able to do return e.g. `["pcdswidgets/ui/vacuum/pump_screens/detailed.ui"]` and have the parent class behavior implement the actual finding of the screens.

Maybe we need to move https://github.com/pcdshub/pcdswidgets/blob/800222b07c4d21300eb32722f1d787917c61c9ef/pcdswidgets/builder/designer_widget.py#L207-L216 into a top-level module and re-use it.

pcdswidgets/vacuum/pumps.py
```diff
@@ -99,9 +116,106 @@ def __init__(self, parent=None, **kwargs):
...
+        controller_base = ""
+        controller_pv = f"{prefix}:VPCNAME"
+
+        try:
+            from epics import caget
+        except ImportError:
+            logger.warning(f"pyepics is unavailable; leaving controller_base empty for {controller_pv}")
+        else:
+            try:
+                controller_value = caget(controller_pv, timeout=1.0)
```
ZLLentz: Suggestion: this shouldn't be done with an operation that can time out, or else the screen freezes sometimes, so it should either change to a value we collect much earlier in e.g. a background thread or with pydm channel, or it should be a static configurable.

ZLLentz (commented):
Some nitpicks, suggestions, and ruminations about splitting the ui folder into "please generate things" and "please do not generate things" halves.

ZLLentz:
The pip builds in the CI here are failing in the same way they did for pcdsdevices... it's frustrating.
We can ignore it for this PR since it's not related, but this makes me want to change the repo to using pixi to run the CI instead of pip or conda.

pcdswidgets/vacuum/base.py
TrumanPar: Oops I'll get them added

pcdswidgets/vacuum/base.py
TrumanPar: I will implement this in the base class. For now, I'll let it resolve to the 'please generate' ui path and I'll change it once we decide on where these should live.

pcdswidgets/vacuum/pump_screens/PIPCombined_controller.py
TrumanPar: No I didn't intend for this, I just haven't spent time figuring out how to work around it. I'll give it some thought and investigation today.

pcdswidgets/vacuum/pumps.py
TrumanPar: I like the idea, for now I'll import it into `vacuum/base.py` and try to cook up an implementation of `get_expert_ui_paths` in the base class instead.

pcdswidgets/vacuum/base.py
TrumanPar: I've decided just to remove the return type to keep consistent with the rest of the file.

pcdswidgets/vacuum/pump_screens/PIPCombined_controller.py
ZLLentz: I think we probably just need to have two ui folders- one for the stuff to generate, and one for static files

pcdswidgets/vacuum/mixins.py
```diff
@@ -59,6 +59,15 @@ def create_channels(self):
...
         if not self._interlock_suffix:
+            if self.interlock_channel is not None:
+                self.interlock_channel.disconnect()
+                self.interlock_channel = None
+            self._interlocked = False
+            self._interlock_connected = False
+            if hasattr(self, "controls_frame"):
+                self.controls_frame.setEnabled(True)
+            if hasattr(self, "update_stylesheet"):
+                self.update_stylesheet()
```
KaushikMalapati: Could we just call self.interlock_value_changed(-1) here?

KaushikMalapati:
> I think the normal thing to do here would be to have separate classes with and without the mixin

I agree with this

pcdswidgets/vacuum/pumps.py
KaushikMalapati: Maybe you can also not show the controller expert screen so we avoid displaying disconnected pvs

pcdswidgets/vacuum/base.py
```diff
@@ -588,6 +588,89 @@ def assemble_layout(self):  # noqa: C901
...
+            local UI files are defined for this expert class.
+        """
+        return []
+
+    @staticmethod
+    def _format_macros(macros: dict[str, str]) -> str:
+        """Serialize a macro mapping into a PyDM-compatible macro string."""
+        return ",".join(f"{key}={value}" for key, value in macros.items())
+
+    def get_expert_macros(self, expert_key, prefix) -> dict[str, str]:
```
KaushikMalapati: Why does this method need expert_key?

pcdswidgets/vacuum/mixins.py
TrumanPar: this works!

pcdswidgets/vacuum/base.py
TrumanPar: I figured that if we do go with the pattern that one widget can support several ophyd classes (it seems like we may just want to make an entirely new widget anyways), the macros might be different. We could always just reference `self.EXPERT_OPHYD_CLASS` so I suppose it is unnecessary.

TrumanPar:
I've pushed a few changes.
- Created a new class `IonPumpNoIlk` which is a copy paste of the regular `IonPump` class minus the interlock mixin. `IonPump` inherits from it and implements interlock mixin plus changes the expert ophyd class + suffixes.
- I've moved `get_expert_ui_paths` implementation to the base class and I use `fix_pcdswidgets_filename`, it seems to be working. I've added a new attribute `EXPERT_UI_DIR`, which is set in the subclass to point to the relative path. Also added a `EXPERT_UI_ORDER` for subclasses to configure the tab order.
- The class doesn't rely on caget called in `get_expert_macros` to get the controller name, I attach it to a channel now.

One side effect of creating a 'IonPumpNoIlk' base class is that we lose the stylesheets defined in lcls-pydm-vacuum, so we would need to include this widget to that repo as well. Otherwise, things appear to be working.

ZLLentz:
> we lose the stylesheets defined in lcls-pydm-vacuum

this almost makes me annoyed enough to finally do away with the stylesheets, or incorporate them into this repo

KaushikMalapati (dismissed):

TrumanPar:
Before we merge, I'll present at Eng gen for feedback on how the screens look.

carogee:
Sorry I didn't see this PR earlier. I have written some [guidelines for font/colors](https://confluence.slac.stanford.edu/spaces/PCDS/pages/683344859/ECS+UI+UX+Screen+Guidelines+and+Best+Practices#ECSUI%2FUXScreenGuidelinesandBestPractices-WidgetGuidelines%3A) on confluence so that our GUI screens will be somewhat standardized in ECS. Can you see if your screen looks ok according to those standards (I definitely need to publicize them more)? If you have comments on the standards I'm also open to discussion, but just trying to pick _something_ that we can all make consistent-ish screens.

Something I did not think about in the guidelines is the tabs--I think elsewhere we have horizontal tabs only, so my gut is to go with horizontal tabs, unless you had a strong feeling for the vertical tabs here?

pcdswidgets/static_ui/vacuum/pumps/PIPCombined_controller.ui
ZLLentz, line 1: Comment:

I think `static_ui` is a reasonable place to store these
Maybe later I will rename both `ui` and `static_ui` to make everything fully explicit- I'm not sure. `ui/static` and `ui/dynamic` maybe? A future question.

pcdswidgets/vacuum/base.py
```diff
@@ -588,6 +596,108 @@ def assemble_layout(self):  # noqa: C901
...
+
+        Returns
+        -------
+        list[str]
+            Absolute paths to the discovered UI files.
+        """
+        if not self.EXPERT_UI_DIR or not expert_key:
+            return []
+
+        from ..builder.designer_widget import fix_pcdswidgets_filename
```
ZLLentz: Change request: we shouldn't be doing inline imports. Can this be moved to the top of the file?
Please take no offense when I say this is an LLM indicator- I'll need to review the rest more closely than usual because of it.

TrumanPar:
That's on me for not adding you to the PR, sorry Carolyn. I think the existing behavior for vacuum widgets using the ui_file_paths designer property is vertical tabs, but we can make it horizontal. I can't think of a good reason for one vs the other but so long as we're consistent right?

pcdswidgets/vacuum/base.py
TrumanPar: I was going to import at the top but copilot suggested to move it down to reduce overhead when we open screens. We do this with typhos as well, but I figure we do it in `_handle_icon_click` since it actually introduces overhead plus in case it raises an error.

ZLLentz: Copilot is wrong much of the time, there is no conceivable overhead from this import

ZLLentz: The typhos import is embedded because it's demonstrably slow (and also optional)

pcdswidgets/vacuum/base.py
```diff
@@ -626,35 +736,46 @@ def _handle_icon_click(self):
...
+        self.tab_widget.setTabPosition(QTabWidget.TabPosition.West)
 
-        kwargs = {"name": name, "prefix": prefix}
-        display = typhos.TyphosDeviceDisplay.from_class(klass, **kwargs)
-        self._expert_display = display
-        display.destroyed.connect(self._cleanup_expert_display)
+        for spec in self.get_expert_tab_specs(klass, prefix):
+            embedded = PyDMEmbeddedDisplay()
+            embedded.set_macros_and_filename(spec["path"], spec["macros"])
+            self.tab_widget.addTab(embedded, spec["title"])
```
ZLLentz, line 748: Side-thought: is it worth being more defensive here? Since we're throwing around a basic dict there's no guarantee about what keys exist and such. Maybe we don't care so much because it has to be defined in this module anyway?

pcdswidgets/vacuum/pumps.py
```diff
@@ -99,9 +109,137 @@ def __init__(self, parent=None, **kwargs):
...
+        Callback invoked when the value changes for the Controller Channel.
+        This callback updates the cached controller base name used by the expert screen.
+
+        Parameters
+        ----------
+        value : int
+        """
+        if value is None:
+            return
+        self._controller_base = value
```
ZLLentz, line 136: Nitpick: there's some inconsistent typing here- `self._controller_base` can be either an empty string or an int. I suggest that it's good practice to be consistent with typing when we can.

pcdswidgets/vacuum/pumps.py
```diff
@@ -99,9 +109,137 @@ def __init__(self, parent=None, **kwargs):
...
+    def get_expert_macros(self, prefix):
+        """
+        Provide expert-screen macros for IonPump.
+
+        Subclasses can tailor this further for IOC naming differences.
+        """
+        macros = super().get_expert_macros(prefix)
+
+        if not self._controller_base:
+            logger.warning(f"No controller base available for IonPump expert macros ({prefix})")
```
ZLLentz, line 150: Comment: this is a good solution to avoid the previous caget timeout issue :+1:

ZLLentz:
I threw in my suggestions and nitpicks- I only ask that you consider the suggestions and complete the pre-merge checklist. It seems pretty solid.

pcdswidgets/vacuum/pumps.py
TrumanPar: Good catch, my laziness in copy pasting from the mixins is to blame. This should always be a string.

pcdswidgets/vacuum/base.py
ZLLentz: I realized offline that the typhos import isn't actually slow in this context due to how pydm loads plugins- we pay the import price early no matter what we do

ZLLentz:
> The pip builds in the CI here are failing in the same way they did for pcdsdevices... it's frustrating.

This fixed itself at some point- I assume some upstream issue was resolved.

pcdswidgets/vacuum/base.py
TrumanPar: maybe instead of wrapping this in a try-except for `KeyErrors`, we add something like:
` if spec.get("path") is None:
      continue` ?
But I feel like if we want to add some checks for valid ui paths, we should do the same for `self.ui_file_paths`, considering this is a designer property and it's easy to make a typo.

TrumanPar: I figure just adding the key check is sufficient. Then again, so long as `get_expert_tab_specs` isn't overriden in the subclass, we are guaranteed to always have the path key.

ZLLentz: Yeah I'm 50/50 on whether anything needs to be done at all- maybe this is something we'd usually pick up via unit tests?

TrumanPar:
@carogee I'm changing screens to adhere to the style sheets in the linked confluence page, here's what they're looking like now:

image

ZLLentz:
Is it possibly a good time (in a separate PR) to bring the functionality from the stylesheet into this repo? It's always bugged me that they were so separate. It's coming up now because of the noilk version not having a stylesheet entry.

I'm thinking either setting a default stylesheet property on each individual vacuum widget or changing the behavior to set the colors etc. not by stylesheet.

I'm happy to tackle this myself if it seems like a good idea, or it's open season if anyone else is more excited to do it.

TrumanPar:
I think moving the stylesheets into this repo and setting a default per widget makes sense. I think it would also be super useful if we add the functionality to give users a drop down selection in the 'view' tab to switch between the standard and a laser-goggle-friendly option as well that can apply globally to the screen. I've heard discussions about this in the past so maybe it's a good time to roll that out too?

TrumanPar:
I believe the pre-merge checklist is complete

pcdswidgets/vacuum/pumps.py
KaushikMalapati: @TrumanPar I think this and the comment about are still outstanding?

TrumanPar: I've updated the doc strings to specify `int` as the return type instead of `string`. Do you think this is worth to add type hinting?

KaushikMalapati: I don't think it needs type hints (also controller_value_changed doesn't return anything). I think Zach was pointing out that self._controller_base is set to the empty string in __init__ and create_channels, but can be set to int in controller_value_changed, changing type from string to int. It doesn't cause an issue here because the not check on an empty string will return True, but maybe _controller_base could be set to None instead of "" to clarify that it never has a string value

TrumanPar: yeah, brain fart moment. I meant to say the input param is now set to a `string` in the doc strings now (not an `int`) but yes this is besides the point. I should spend more time thinking / revisiting the code before I type. 

Regardless, I see that I actually never pushed that change and I don't know why. The doc strings should specify a string. I suppose if we want to stop the input from being an int, we can do type hints, no?

KaushikMalapati: type hints won't affect anything during runtime (unless you use specific tools to make that happen), they only communicate what is supposed to happen and can also help with linting. If it supposed to be a string, then changing the `value : int` to `value : str` is sufficient here so that we know that controller_value_changed expects value to be an int (when I was writing my previous comment I thought it was supposed to be an int because I only read the parameters part of the docstring)

TrumanPar: Wow, I don't know how I've gone this long thinking type hints enforced types. But sounds good, I'll push that change

KaushikMalapati (approved):

KaushikMalapati (approved):

TrumanPar:
@KaushikMalapati I had to pull in from master before merging, could you approve again? Thanks

KaushikMalapati (approved):

ZLLentz:
There weren't any designer screenshots included

TrumanPar:
I thought a screenshot of the widget loaded in designer was sufficient. Do you intend for me to take a screenshot of the widget discovered/ loaded into the sidebar? Happy to do so

ZLLentz:
Is the first screenshot, labelled "loaded in pydm", actually loaded in designer?

TrumanPar:
Ah, I think that screenshot was taken from a pydm instance. I've included a picture actually in designer.

ZLLentz:
Thank you! And great work on the PR!

PR #116, opened 2026-06-22
Cam Stream Widget:  Marker Overlay and Minor Enhancements

ipotanin:
## Description

- Added "marker_selection" sub widget for selection of points tied to EPICS PVS (crosshairs)
    - this includes a CamMarker object that keeps track and draws each crosshair on screen
    - added a marker style dialog, to change visual properties of this marker (with, line pattern, length)
- Added macros for specifying the full Stream PV ( protocol name, stream name, and suffix) with sensible defaults
     - This allows you to view an ROI stream or fully define the PV used to get the image stream
- Updated all camera sub-widget icons for designer
- made sub-widget name inference more explicit (if nickname property is defined, use that)
- FIX: row/column order was wrong the whole time, this has been fixed

## Motivation and Context

- this is part of development of laser pointing gui for the ASC lab
- lab uses a lot of markers to keep track of points of interest
- camviewer-like features are added to pcdsdevices to be re-used in other projects

## How Has This Been Tested?

-interactively
- tested with several cameras in ASC lab
- tested viewing an ROI stream on one tab and a full view stream on the other. updating ROI in the main view works as expected

here

## Screenshots

image
image

image

## Pre-merge Checklist


ipotanin:
Known issues/quirks not addressed in this PR but hope to address in future:

- All ROIs and Markers are defined in screen pixels. If the stream has binning or start offset on any upstream channel the markers will not match. Still thinking of ways to have unit modifiers on these, so if you are viewing a binned or cropped stream, the X,Y points are still recorded to the PVs in full resolution pixel counts. 
- pyDMImage (or any pcds composite widget) will always first try to connect to the non-substituted PV names (PV:${name}:TEST) before timing out. This occasionally results in 30seconds of no frames after start-up.  I do not have a good idea in mind for the fix for this.

ZLLentz:
From the PR description: I like how this looks a lot

ZLLentz:
> pyDMImage (or any pcds composite widget) will always first try to connect to the non-substituted PV names (PV:${name}:TEST) before timing out. This occasionally results in 30seconds of no frames after start-up. I do not have a good idea in mind for the fix for this.

Hopefully we can solve this centrally in the builder scripts to avoid ever having an unsubstituted macro applied

pcdswidgets/imaging/common/collapsible_section.py
```diff
@@ -48,9 +48,9 @@ def __init__(
     @staticmethod
     def _infer_title(content: QWidget) -> str:
         """Extract a display title from a widget."""
-        nickname = getattr(content, "nickname", None)
-        if isinstance(nickname, str) and nickname.strip():
-            return nickname.strip()
+        nickname = getattr(content, "nickname", None) or content.property("nickname")
```
KaushikMalapati:
```suggestion
        nickname = getattr(content, "nickname", content.property("nickname"))
```

carogee:
It looks very nice! Are the cross hair colors chosen by the user (They are different in the two images)? The red/green combination is a common color-blindness challenge. Also may want to check colors with laser goggles if you haven't already and if this will be commonly used by laser folks.

pcdswidgets/imaging/common/collapsible_section.py
```diff
@@ -48,9 +48,9 @@ def __init__(
...
+        if isinstance(nickname, str) or content.property("nickname") and nickname.strip():
+            return nickname.strip() or content.property("nickname")
```
ZLLentz: Question: what's the failure mode of an unfilled nickname property? Does it just return e.g. empty string?

pcdswidgets/imaging/common/marker_selection_full.py
```diff
@@ -0,0 +1,287 @@
...
+from pcdswidgets.imaging.common.marker_style_dialog import MarkerStyleDialog
+
+logger = logging.getLogger(__name__)
+
+# Default colors per marker index (these can be overwritten in designer)
+_DEFAULT_COLORS = [
+    QColor("red"),
+    QColor("green"),
+    QColor("cyan"),
+    QColor("magenta"),
```
ZLLentz: Comment: camviewer used blue, red, cyan, magenta
I think your colors are probably better (having both blue and cyan seemed weird to me)

ZLLentz: Based on the comment from @carogee maybe we ought to avoid red/green as the defaults

ipotanin:
> Are the cross hair colors chosen by the user (They are different in the two images)? The red/green combination is a common color-blindness challenge. Also may want to check colors with laser goggles if you haven't already and if this will be commonly used by laser folks.

Yes, user can set color on runtime( a selection diagram comes up when clicking on the color):
image

You can override defaults in designer 
image


Do you have suggestions for the default-defaults?

ZLLentz:
Side comment: I like the new icons, it's easy and intuitive which is the main widget and which are the setting extension widgets when scanning through the designer sidebar

carogee:
Screenshot 2026-06-25 at 3 17 05 PM
Great that the colors are chosen by the user! It looks like the colors will be fairly distinguishable if the red or green is replaced by yellow. So maybe green, yellow, cyan, and magenta by default? I'm not familiar with which colors are hard to see with the laser goggles though.

pcdswidgets/imaging/common/marker_selection_full.py
```diff
@@ -0,0 +1,287 @@
...
+        self._apply_default_colors()
+        self._connect_buttons()
+        self._connect_spinboxes()
+
+    def _set_macro_defaults(self):
+        """Populate unset macros with sensible defaults."""
+        default_map = {}
+        for i in range(NUM_MARKERS):
+            default_map[f"suffix_{i + 1}x"] = f":Over1:{i + 5}:PositionX"
+            default_map[f"suffix_{i + 1}y"] = f":Over1:{i + 5}:PositionY"
```
ZLLentz, line 78: Comment: camviewer used `:Cross1X`, `:Cross1Y`, `:Cross2X` etc. to avoid editing plugin PVs- should we do the same here or is there some value in linking to the overlay plugin? Maybe not using the save PVs as camviewer is actually preferrable? Any thoughts?

pcdswidgets/imaging/common/collapsible_section.py
ipotanin: I didn't pay enough attention to this, and I think this is some slop fix that I did not vet properly. 

It looks like ` getattr(content, "nickname", None) `and `content.property("nickname")` are equivalent statements.

ZLLentz:
(The widget itself works great, I took it for a spin)

pcdswidgets/imaging/common/marker_selection_full.py
ipotanin: So I learned the fun way that CROSS1X is linked to IMAGE1:Overlay plugin in camviewer.

I explicitly chose Over1 in case we move away from stream specific plugins (works with vanilla area detector),  but also works with our common gige

these are defaults that anyone can override, I'm open to suggestions too

pcdswidgets/imaging/common/collapsible_section.py
ipotanin: both return NONE if no property is set.

The behavior evaluates as desired, but I can and should clean this up.

pcdswidgets/imaging/common/marker_selection_full.py
ZLLentz: I think it's a good choice- I wanted to highlight that it was different, make sure it was deliberate, and document it in a PR discussion.

ZLLentz (approved):
I'm :+1: on this
I made sure I had a self-assigned ticket for the bad-value-at-init race condition, which affects all the widgets but is more consequential with image PVs

PR #117, opened 2026-06-24
CI: GHA SHA test in advance of enterprise requirements

ZLLentz:
## Description

Update the pcds-ci-helpers reference to use a slightly older SHA (to see if dependabot will update it for us)
Add a dependabot config to keep the SHA periodically synchronized

## Motivation and Context

At the start of July this will be required or else all GHA builds will no longer run
https://jira.slac.stanford.edu/browse/ECS-10557

I want to get this working manually on a few repositories before trying to automate the update everywhere
In pcdswidgets, I will check that setting to an older SHA "just works" and that dependabot will automatically update it.

I'm picking a monthly update so we don't get crazy spam PRs from dependabot too often, but we still stay up-to-date.
A repo that needs an immediate update can force dependabot to run early via the github ui.

## How Has This Been Tested?

I tested dependabot lightly over in pcds-ci-helpers
I'm planning to use the results of this CI run as a green light that this is the correct way to do an update

Here only

## Pre-merge Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    cooldown:
+      default-days: 7
+      exclude:
+      - pcdshub/pcds-ci-helpers/*
--- a/.github/workflows/standard.yml
+++ b/.github/workflows/standard.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   standard:
-    uses: pcdshub/pcds-ci-helpers/.github/workflows/python-standard.yml@master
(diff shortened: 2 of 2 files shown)


KaushikMalapati (approved):

KaushikMalapati:
@ZLLentz how does one manually run dependabot? Is this an admin-only power?

ZLLentz:
I'm not sure if it's admin-gated, but it's certainly obfuscated and it does require this dependabot.yml file to exist.

- Click the top-bar "insights" tab
- Click the left-side "dependency graph" option
- Click the middle-top "dependabot" tab

ZLLentz:
I think there's also a way to trigger dependabot updates for security advisories through the top security tab, which doesn't require any specific repo config

PR #122, opened 2026-07-07
ENH: port the lucid toolbar widget with only minor changes

ZLLentz:
## Description

Bring the `QuickAccessToolbar` from `lucid` into `pcdswidgets`, renaming it to `YamlToolbar`.

I've made only minimal changes related to code quality and related to enabling smooth usage in designer, without adding anything beyond that. The changes I've made in roughly the order you'll encounter them in the file are:

- Add a `pyyaml` dependency, which was a dependency of `lucid`
- Add a module docstring
- Point imports to pcdswidgets instead of to lucid
- Rename `QuickAcessToolbar` to `YamlToolbar`
- Add a better docstring to `YamlToolbar`
- Configure `YamlToolbar` for designer inclusion
- Add type annotations
- Remove functions that are only called once
- Make the filename property exist
- Make it so we don't get spammed with errors in designer as we type into the filename property setter
- Stop passing around instance variables when not needed, use function args instead
- Allow the load to continue if individual buttons are misconfigured
- Add missing docstrings to various methods
- Rename some confusingly-named variables
- Have proper error handling for incorrect button configurations instead of making a non-functional button
- Rename `SnakeLayout`, which does not involve snaking, to `YamlTabLayout`, as it was only ever used here.
- Include the layout in the same module file instead of elsewhere in the library.
- Fix errors in the layout's docstring
- Add the only test that was present in lucid for this code
- Add the new widget to designer via `make pyproject.toml`

Some future changes I'm considering for follow-up PRs are:

- [ ] Add a helper to generate a ui file from a lucid yaml file, so people can edit these as ui files instead of as yaml (I know there are conflicting opinions on this)

## Motivation and Context

This is part of https://jira.slac.stanford.edu/browse/ECS-10760
Where we'd like to be able to use the tools from lucid in arbitrarily designed pydm home screens instead of locking them all to some fixed configurations.

## How Has This Been Tested?

I added the layout test that had been previously included in lucid, but I did not write a new test for the widget itself.
I'm testing interactively one more time before opening for review in order to collect screenshots

Here only

## Screenshots

In designer: you get to see the buttons and tabs load in. You can browse the tabs but you cannot click the buttons.
image

In pydm: it looks and acts as you'd expect
image

## Pre-merge Checklist

### DIFF
--- a/pcdswidgets/common/toolbar/yaml_toolbar.py
+++ b/pcdswidgets/common/toolbar/yaml_toolbar.py
@@ -1,57 +1,100 @@
+"""
+A button toolbar loaded from a lucid-style yaml file.
+
+This was originally part of lucid but has moved here so it can be used without
+the lucid launcher, of which it has no real dependencies.
+"""
+
 import logging
 import os
+from typing import Any
 
 import yaml
 from pydm.widgets import PyDMRelatedDisplayButton, PyDMShellCommand
 from qtpy import QtCore, QtWidgets
 from qtpy.QtCore import Qt
-from qtpy.QtWidgets import QGridLayout, QPushButton
+from qtpy.QtWidgets import QGridLayout, QPushButton, QWidget
 
 from pcdswidgets.common.dock.tab_dock_button import TabDockButton
+
+try:
+    from qtpy.QtCore import Property  # type: ignore
+except ImportError:
+    from qtpy.QtCore import pyqtProperty as Property  # type: ignore
 
 logger = logging.getLogger(__name__)
 
 
-class QuickAccessToolbar(QtWidgets.QWidget):
-    """Tab Widget with tabs containing buttons defined via a yaml file"""
+class YamlToolbar(QtWidgets.QWidget):
+    """
+    Tab Widget with tabs containing buttons defined via a yaml file.
 
-    def __init__(self, parent=None):
+    This uses the legacy yaml format from lucid, which defines which buttons go in which tab.
+
+    See https://pcdshub.github.io/lucid/master/toolbar.html for more information.
+    """
+
+    _qt_designer_ = {
+        "group": "ECS Common Toolbar",
+        "is_container": False,
+    }
+
+    def __init__(self, parent: QWidget | None = None):
         super().__init__(parent=parent)
 
-        self._tools_file = ""
-        self._tools = None
+        self._config_file = ""
         self._default_config = {"cols": 4}
         self.default_dock_button = None
-        self._setup_ui()
 
-    def set_tools_file(self, file: str):
-        if not file:
-            return
-        self._tools_file = file
-        with open(file) as tf:
-            self._tools = yaml.full_load(tf)
-        self._assemble_tabs()
-
-    def _setup_ui(self):
         self.setSizePolicy(QtWidgets.QSizePolicy.Preferred, QtWidgets.QSizePolicy.Preferred)
-
         main_layout = QtWidgets.QVBoxLayout()
         self.setLayout(main_layout)
         self.tab = QtWidgets.QTabWidget()
         main_layout.addWidget(self.tab)
 
-    def _assemble_tabs(self):
-        if self._tools is None:
+    def get_config_file(self) -> str:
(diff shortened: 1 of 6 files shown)


pcdswidgets/common/toolbar/yaml_toolbar.py
```diff
@@ -0,0 +1,194 @@
...
+            config.update(tab_params.get("config", {}))
+
+            cols = config.get("cols", 4)
+            page.setLayout(YamlTabLayout(cols))
+
+            buttons = tab_params.get("buttons", {})
+            for button_text, button_config in buttons.items():
+                try:
+                    button_widget = self._button_factory(button_text, button_config)
+                except Exception as exc:
```
KaushikMalapati, line 103: Should this only catch RuntimeException since that's the only things _button_factory directly raises?

pcdswidgets/common/toolbar/yaml_toolbar.py
```diff
@@ -0,0 +1,194 @@
...
+                    logger.error(f"Error creating {button_text} button with exception {exc}, skipping.")
+                    logger.debug(exc, exc_info=True)
+                    continue
+                page.layout().addWidget(button_widget)
```
KaushikMalapati, line 107: I wonder if it's more "pythonic" to use else or continue here

pcdswidgets/common/toolbar/yaml_toolbar.py
```diff
@@ -0,0 +1,194 @@
...
+
+    See https://pcdshub.github.io/lucid/master/toolbar.html for more information.
+    """
+
+    _qt_designer_ = {
+        "group": "ECS Common Toolbar",
+        "is_container": False,
+    }
+
+    def __init__(self, parent: QWidget | None = None):
```
KaushikMalapati, line 42: I'm sure you've seen this, but maybe it's easier to use [typing.Optional](https://docs.python.org/3/library/typing.html#typing.Optional) in the future (not needed for this pr)

KaushikMalapati (approved):
Some nits but no blockers

pcdswidgets/common/toolbar/yaml_toolbar.py
ZLLentz: I went back and forth on this, but eventually settled on catching anything so that the screen can still load "something" when individual buttons have bad entries, for example if a bad value is set to a property.

pcdswidgets/common/toolbar/yaml_toolbar.py
ZLLentz: `typing.Optional` is actually the old way to specify these, it's more compact in some ways but it's a misleading type hint because it re-uses the same word as an "optional" argument, which just implies that you don't need to pass it and might never be `None`, whereas the `Optional` type means something-or-None, so to avoid the inconsistent language this is something I don't use any more.

Usually you would use `typing.Optional` for compatibility with python 3.9 and earlier, which don't have access to the `|` syntax.

pcdswidgets/common/toolbar/yaml_toolbar.py
ZLLentz: I think it's personal preference on this one, I'll explain my reasoning though:

I avoid `try/except/else` when possible because it confuses some people (and me sometimes) since the `else` keyword is most commonly used in `if/else` and the keyword ends up being re-used for a completely different purpose.

Unrelated to `else`, I also try to minimize indents in this language (maybe a bit too dogmatically) because it can become unwieldy.

pcdswidgets/common/toolbar/yaml_toolbar.py
KaushikMalapati: Thanks for letting me know

pcdswidgets/common/toolbar/yaml_toolbar.py
KaushikMalapati: Sounds good

ZLLentz:
Thanks for the review, I'll merge this in and later I'll probably port the lucid grid to finish the set

PR #123, opened 2026-07-09
ENH: port the lucid IndicatorGrid

ZLLentz:
## Description

Add the `IndicatorGrid` from `lucid` and clean it up/tweak it a bit. (Probably more cleanup could be done).

No significant features were added compared to the `lucid` version, but some unused elements were cut.

Major additions/changes:
- Add `IndicatorGrid` and all supporting code/helpers from `lucid`
- Add optional `"typhos"` dependency group that has `happi` and `typhos`, and include them in a second `pixi` env alongside `pcdsdevices` to be used for testing. You can pass the `--environment typhos` option to `pixi` to e.g. run with `typhos` included.
- Add the grid tests that already existed in `lucid`
- Significant rewrite of docstrings and type annotations for these classes and methods
- Change `IndicatorGrid` so that it can be included at all using `designer`
- Embed some of the `lucid` launcher setup into `IndicatorGrid` (only invoked if we use `happi`)
- Reduce some of the error spam when devices can't load into the grid- keep the tracebacks to debug level

Possible follow-up actions for another PR:
- [ ] More cleanup? E.g. my handling of the yaml layout should be centralized now
- [ ] Can we make `IndicatorGrid` look nicer in designer? Would some minimum sizing help us reserve space in the pydm screen too during the loading process?
- [ ] Add options for specifying which parts of happi are or are not included in the grid
- [ ] Add options for manually adding indicator + ui file pairs to the grid (and linking them to dock buttons too?)
- [ ] Improve user experience as grid is filling? Maybe the grid can fill slowly as devices load in instead of mostly all at once?

## Motivation and Context

We'd like to be able to put an entire lucid home screen into a pydm screen for hutch-local optimization- and now we can.
See https://jira.slac.stanford.edu/browse/ECS-10760

## How Has This Been Tested?

Interactively
The unit tests from lucid were added here, but no additional tests were added

Here only

## Screenshots

Note: this requires the typhos module at runtime, hence my usage of the alt environment here

image

But the designer never requires it (since we don't want to slow ourselves down with happi loads in designer anyway, and we want to be able to edit these always). Note that the grid currently is just a tiny spec in designer by default- not sure if I should address it now or wait till the cleanup PRs.

image

## Pre-merge Checklist


pcdswidgets/common/dock/indicator_grid.py
```diff
@@ -0,0 +1,542 @@
...
+            beamline=[beamline], group_keys=("location_group", "functional_group"), callbacks=[self.add_from_dict]
+        )
+        self.loader.start()
+
+    def configure_ophyd(self):
+        """
+        Configure settings for ophyd that make the grid experience smoother.
+
+        For this to work properly, this widget must be the first thing in the application to be using
+        ophyd devices, which should usually be true.
```
KaushikMalapati: What happens if thsi widget isn't the first to use ophyd?

ZLLentz:
Last night I was realizing that I might not want to commit to a `beamline` widget property, since it could become deprecated quickly as options are added. It might be worth brainstorming what the properties should look like "later" and implement some of that now to avoid deprecations etc., not sure though

pcdswidgets/common/dock/indicator_grid.py
ZLLentz: iirc various ophyd configuration steps error out if you've already been creating ophyd devices- let me check on this and confirm behaior, it's worth doing "something" about this (don't crash? better error messages? better docs?)

ZLLentz: My concerns were on these two calls:

- `OphydObject.add_instantiation_callback` I thought at some point errored by default if an ophyd object already existed, but this it no longer the default behavior (it's opt-in today), so this one is safe, see https://github.com/bluesky/ophyd/blob/5a3453c7d2af94dd3fd32d55f5b56c6b742e8f78/ophyd/ophydobj.py#L211-L217
- `EpicsSignalBase.set_defaults` always raises a RuntimeError if any epics signals had already been created, see https://github.com/bluesky/ophyd/blob/5a3453c7d2af94dd3fd32d55f5b56c6b742e8f78/ophyd/signal.py#L1158-L1164

I think based on that research, what I'll do is wrap the `set_defaults` call in a try/except and have a warning show up instead of giving up on the whole setup, and adjust the docstring accordingly.

ZLLentz:
After some thought today, I want to keep `beamline` around indefinitely as a "shortcut" for the beamline=, active=True search. So even if I add more granular search options later, this one could stick around and not conflict with e.g. a `happi_searches` property and whatever manual configuration may be added later to include ui files.

KaushikMalapati (approved):

PR #126, opened 2026-07-15
Migrating Stylesheets

TrumanPar:
## Description

Moving the existing master style sheet, which currently lives in lcls-pydm-vacuum, into this repo. I've also added stylesheets for the new `IonPumpNoIlk` class and I set `PYDM_STYLESHEET` to point to this path in the `env_setup.sh` script.

## Motivation and Context

We want to move stylesheets into this repo such that all dependencies are in one place.

## How Has This Been Tested?

Interactively with the new Ion Pump class.

## Screenshots

In designer with new style sheets:
image

and without the stylesheets:
image

You can see how the new stylesheet has applied. There are no errors upon launching designer or pydm that would indicate the stylesheets are applied incorrectly.

Screenshot from the working widget in pydm (showing both new and existing Ion Pumps)
image

## Pre-merge Checklist


pcdswidgets/stylesheets/masterStyleSheet.qss
KaushikMalapati, line 1: The IonPumpNoIlk style rules look identical to the IonPump style rules, but IonPump is a child of IonPumpNoIlk. Does IonPump inherit its style from IonPumpNoIlk's rules if IonPump didn't have any explicit style rules in this file? Would make it easier to update both in the future

pixi_scripts/env_setup.sh
```diff
@@ -5,3 +5,4 @@ THIS_ENV="${HERE}"/..
 export PYQTDESIGNERPATH="${THIS_ENV}"/.pixi/envs/default/lib/python3.12/site-packages/pydm
 export PYDM_DESIGNER_ONLINE=1
 export QT_XCB_GL_INTEGRATION=none
+export PYDM_STYLESHEET="${THIS_ENV}"/pcdswidgets/stylesheets/masterStyleSheet.qss
```
KaushikMalapati: @ZLLentz do you know if env_setup.sh is called if you do pixi run? Wondering if this is the right place to set the environment variable so that it always get picked up

KaushikMalapati: Also, should this be 
```suggestion
export PYDM_STYLESHEET="${THIS_ENV}"/pcdswidgets/stylesheets/masterStyleSheet.qss:$PYDM_STYLESHEET
```
so that we don't overwrite any other files?

TrumanPar: I believe so according to their docs (also thanks for sharing them with me Kaushik) "This command is used to run tasks in the pixi environment. It will activate the environment and run the task in the environment. It is using the deno_task_shell to run the task" https://pixi.prefix.dev/latest/reference/cli/pixi/run/#global-options

I've tested using `pixi run designer` and `pixi shell`, then running designer (both accomplish the same thing)

pcdswidgets/stylesheets/masterStyleSheet.qss
TrumanPar: After testing, I can confirm it does apply to the derived class as well

pixi_scripts/env_setup.sh
TrumanPar: I can't think of any downsides for this

ZLLentz: Double-confirming that anything in the script there will be used because it's listed as an activation step in https://github.com/pcdshub/pcdswidgets/blob/88705d45575a854ccdbc835c324d43c3ea318705/pixi.toml#L10-L11

Since we're distributing this in a shell script we'll separately need to update the `ctrlenv-widgets` and `ctrlenv-lucid` pixi envs in a similar way (outside of this PR, after the tag, before the pixi env release). I'd like to be able to distribute these without shell scripts at some point in the future.

pixi_scripts/env_setup.sh
```diff
@@ -5,3 +5,4 @@ THIS_ENV="${HERE}"/..
 export PYQTDESIGNERPATH="${THIS_ENV}"/.pixi/envs/default/lib/python3.12/site-packages/pydm
 export PYDM_DESIGNER_ONLINE=1
 export QT_XCB_GL_INTEGRATION=none
+export PYDM_STYLESHEET="${THIS_ENV}"/pcdswidgets/stylesheets/masterStyleSheet.qss:$PYDM_STYLESHEET
```
ZLLentz: Nitpick: consistent variable style
```suggestion
export PYDM_STYLESHEET="${THIS_ENV}"/pcdswidgets/stylesheets/masterStyleSheet.qss:"${PYDM_STYLESHEET}"
```

ZLLentz (dismissed):
There's nothing objectionable here- very succinct way to move the stylesheet over. If we want to integrate it more tightly we can do so later now that it's all in one place.

KaushikMalapati (dismissed):

KaushikMalapati (approved):

PR #133, opened 2026-08-06
Add Motor Tip Tilt Widgets

moyinoluwa-10:
## Description
Adds `MotorTipTiltFull`/`MotorTipTiltDouble`, generalized versions of the existing `SmaractTipTiltFull`/`SmaractTipTiltDouble` widgets so any tip/tilt motor pair (using motor records and related software) can use them.

A new `motor_style` property (`MotorStyle.MOTOR_RECORD` / `MotorStyle.SMARACT`) lets the same widget drive either a normal EPICS motor record or a SmarAct MCS2 controller's step-based PVs, so one widget now covers both cases. Also adds a shared stop button that writes to both axes at once.

Also adds a `set_motors()` helper to the smaract and motor tip/tilt widgets to swap which PVs drive each axis without losing the invert-button state (setting the macros directly would otherwise reset the forward/backward mapping).

Registers `MotorTipTiltDouble` and `MotorTipTiltFull` in `pyproject.toml`.

> [!WARNING]
> **Breaking change:** `SmaractTipTiltFull` and `SmaractTipTiltDouble` are removed entirely (code, `.ui` files, and their `pyproject.toml` designer registration), now that `MotorTipTiltFull`/`MotorTipTiltDouble` cover the smaract case via `motor_style`. Any existing screen with a promoted `SmaractTipTiltFull`/`SmaractTipTiltDouble` widget will need to be repointed to `MotorTipTiltFull`/`MotorTipTiltDouble` with `motor_style` set to `smaract`.

## Motivation and Context
Need a reusable tip/tilt widget for motor pairs that aren't smaract specific (e.g. picomotors), and a single widget that covers both motor styles instead of maintaining separate smaract and motor-record implementations.

## How Has This Been Tested?

- Live tested using a picomotor set up in the ASC lab.
- Live tested using a smaract motor set up in TMO.

Docstrings on the new classes and methods.

Designer (widgets appear under the "ECS Motion Common" group):

```
./try_in_designer.sh pcdswidgets/ui/motion/common/motor_tip_tilt_double.ui
```

```
./try_in_designer.sh pcdswidgets/ui/motion/common/motor_tip_tilt_full.ui
```

PyDM:

```
./try_in_pydm.sh -m '{"vertical_motor": "LASDEV:CNT:PIC:03:03", "horizontal_motor": "LASDEV:CNT:PIC:03:04"}' pcdswidgets/ui/motion/common/motor_tip_tilt_double.ui
```

```
./try_in_pydm.sh -m '{"vertical_motor": "LASDEV:CNT:PIC:03:03", "horizontal_motor": "LASDEV:CNT:PIC:03:04"}' pcdswidgets/ui/motion/common/motor_tip_tilt_full.ui
```

## Screenshots

**Designer:**
Screenshot 2026-08-06 at 6 22 05 AM <br/>
Screenshot 2026-08-07 at 10 54 21 AM<br/>
Screenshot 2026-08-07 at 10 55 11 AM

**PyDM:**
- `MotorStyle.MOTOR_RECORD`
Screenshot 2026-08-07 at 8 37 59 AM<br/>
Screenshot 2026-08-07 at 8 38 23 AM

- `MotorStyle.SMARACT`
Screenshot 2026-08-07 at 10 22 07 AM <br/>
Screenshot 2026-08-07 at 10 22 15 AM

## Pre-merge Checklist
- [ ] Description and screenshot and of these widgets working are added to the ECS Widget Catalog https://confluence.slac.stanford.edu/spaces/PCDS/pages/716925985/ECS+UI+UX+pcdswidgets+Widget+Catalog


pcdswidgets/motion/common/motor_tip_tilt_double.py
ZLLentz, line 1: I think we've crossed some sort of code duplication threshold. We have four python files now (up from two) with nearly identical code. This was called out in the previous tip/tilt review as a "let's handle it later, there's just two right now". Is there any way to consolidate this logic into a shared base class or helper functions to share among the four widgets?

aberges-SLAC: I was discussing this with Moyin and I think the right way forward is to:

1. Incorporate the logic in `SmarActTipTilt` into these generic Motor files ✅ 
2. Add a `pyqtProperty` that specifies if this is `Motor Record` or `SmarAct` style
3. Read the property and assign the PV's accordingly:
    * `.TWV` == `:STEP_COUNT`
    * `.TWF` == `:STEP_FORWARD`
    * `.TWR` == `:STEP_REVERSE`
    * `.RBV` == `:TOTAL_STEP_COUNT`
    * `.STOP` can stay `.STOP` obviously (this will be incorporated soon)
 4. Delete the `SmarAct` TipTilt main widgets. The support screens can still exist

Life will be harder when I try to do closed loop motion behaviors in `SmarAct` but that's a future me problem.

ZLLentz: It seems pretty reasonable, but you have to pause for a moment and notice that doing:

> 4. Delete the SmarAct TipTilt main widgets. The support screens can still exist

Will break any existing screen using those widgets. Is that OK? If it isn't, we could come up with some backwards compatibility idea, or we could take a different approach.

aberges-SLAC: I think in this case I would advocate for the API break and wag my finger at folks who deployed off of `latest` instead of using a static, stable release of `ctrlenv-widgets`/`pcdswidgets`.

Besides I'm like the only person deploying this widget beyond maybe Vincent or a handful of screens. 

It's cleanest to nip this in the bud and collapse to the a single set files instead.

moyinoluwa-10: I've made all required updates and updated the PR description to match.

aberges-SLAC (dismissed):
This also looks good to me.

@ZLLentz Do you agree with my heavy-handed solution for SmarActs? I think if we add an appropriate **API BREAK** tag on the next release, it should be sufficient for folks. I think only a handful of us actually deployed this little widget anyhow.

ZLLentz:
@aberges-SLAC my only objection is that you're eating into my unlimited supply of major version update tag numbers

(by the semver rules, we'll need to tag v2.0.0 instead of v1.6.0)

pcdswidgets/motion/common/motor_tip_tilt_double.py
```diff
@@ -0,0 +1,220 @@
...
+                logger.warning(f"Invalid motor_style {value!r}; expected one of {[s.value for s in MotorStyle]}")
+            return
+        if style == self._motor_style:
+            return
+        self._motor_style = style
+        self._apply_expert_screen_visibility()
+        self._refresh_axis("vertical")
+        self._refresh_axis("horizontal")
+
+    motor_style = pyqtProperty(str, getMotorStyle, setMotorStyle)
```
ZLLentz: I forget the details, but there's some horrible thing you can do to get these enums to show up in the designer gui. Consider digging into this an optional suggestion.

pcdswidgets/motion/common/motor_tip_tilt_full.py
ZLLentz, line 1: Were we intending to finish deduplicating the code or was that out of scope?

pcdswidgets/motion/common/motor_tip_tilt_double.py
moyinoluwa-10: Yes, there is. I tried looking into it, but I hit a blocker. It's also good to note that implementation varies between PyQt5 and PyQt6. If I were to implement this, it would need to be updated if we ever switch.

pcdswidgets/motion/common/motor_tip_tilt_full.py
moyinoluwa-10: I think that's out of scope for now. @aberges-SLAC can comment more on that.

ZLLentz:
I'm cloning the repo to do a final qa step and then I'm putting my approval in if it looks normal

ZLLentz (approved):
I think this should be good to go

pcdswidgets/motion/common/motor_tip_tilt_full.py
aberges-SLAC: I'll PR on de-deplicating the code, since it'll prompt me to go fix all my deployments

PR #134, opened 2026-08-06
Add Imaging Centroid Tracker

moyinoluwa-10:
## Description
Adds a new `CentroidTrackerFull` widget for tracking a beam centroid across one or two `PyDMImageView`s (e.g. full view/cropped view) at once, with a threshold that can follow the image's live intensity, a dedicated marker style dialog, and settings that can be saved/restored. 

Other changes needed along the way:
- `CamMarker`/`MarkerStyleDialog` get two new marker style and a radius-style option, and can now draw onto two `PyDMImageView`s at once with configurable offsets.
- `ColormapIntesityControlFull` can drive two image views and remembers its saved levels.
- Adds a shared "save settings" signal/helper used by the ROI, marker selection, collapsible section, and centroid tracker widgets.
- Registers `CentroidTrackerFull` in `pyproject.toml`.

## Motivation and Context
This came out of alignment work that needs to track a beam centroid on two camera views (near-field/far-field) at the same time.

## How Has This Been Tested?

Live tested using a camera setup in TMO.

Docstrings on the new/changed classes and methods. 

Designer (widget appear under the "ECS Imaging Common" group):

```
./try_in_designer.sh pcdswidgets/ui/imaging/common/centroid_tracker_full.ui
```

PyDM:

```
./try_in_pydm.sh -m '{"cam_prefix": "LM1K4:EJX_DP0_TF1_NF1", "stat_plugin": ":Stats2:", "suffix_centroid_x": "CentroidX_RBV", "suffix_centroid_y": "CentroidY_RBV", "suffix_sigma_x": "SigmaX_RBV", "suffix_sigma_y": "SigmaY_RBV", "suffix_max_value": "MaxValue_RBV","suffix_centroid_threshold_rbv": "CentroidThreshold_RBV"}' pcdswidgets/ui/imaging/common/centroid_tracker_full.ui
```

## Screenshots

**Designer:**
Screenshot 2026-08-06 at 11 29 04 AM
Screenshot 2026-08-06 at 11 38 51 AM

**PyDM:**

Screenshot 2026-08-06 at 11 57 37 AM
Screenshot 2026-08-06 at 11 59 01 AM

**Marker Styles:**
- CROSSHAIR_LENGTH
Screenshot 2026-08-06 at 3 58 11 PM

- INFINITE_LINES
Screenshot 2026-08-06 at 3 58 19 PM

- ELLIPSE
Screenshot 2026-08-06 at 3 59 43 PM

- INFINITE_LINES_AND_ELLIPSE
Screenshot 2026-08-06 at 3 59 22 PM

## Pre-merge Checklist
- [ ] Description and screenshot and of these widgets working are added to the ECS Widget Catalog https://confluence.slac.stanford.edu/spaces/PCDS/pages/716925985/ECS+UI+UX+pcdswidgets+Widget+Catalog


pcdswidgets/imaging/common/cam_marker.py
```diff
@@ -1,25 +1,34 @@
...
+_ELLIPSE_SEGMENTS = 64
+
 
 class MarkerStyle(IntEnum):
     """Available marker display styles."""
 
     CROSSHAIR_LENGTH = 0
     INFINITE_LINES = 1
+    ELLIPSE = 2
+    INFINITE_LINES_AND_ELLIPSE = 3
```
ZLLentz, line 20: Optional suggestion: should we separate "shape" into a separate enum? Or maybe I'm misunderstanding what this does here. It seems to me like this enum will become burdensome as we add different marker shapes (e.g. triangle, infinite_lines_and_triangle). I'm curious for any thoughts on this, maybe we look into it if we expand the enum again.

pcdswidgets/imaging/common/cam_marker.py
```diff
@@ -1,25 +1,34 @@
...
 
 class CamMarker:
     """Composite marker overlay for a single point of interest.
 
-    Renders on a pyqtgraph ViewBox as either crosshairs of varying sizes
-    or infinite (full-span) lines.  Wraps the graphic items so style
-    changes preserve the marker position.
+    Renders on one or more pyqtgraph ViewBoxes as crosshairs of varying
+    sizes, infinite (full-span) lines, an ellipse, or infinite lines
+    combined with an ellipse. Wraps the graphic items so style changes
```
ZLLentz, line 28: For documentation: do you have a screenshot to share for these new marker styles? It'd be useful to include this in the PR description.

pcdswidgets/imaging/common/cam_marker.py
```diff
@@ -29,6 +38,17 @@ class CamMarker:
         Initial pen width in pixels.
     style : MarkerStyle
         Initial display style.
+    arm_length : int
+        Initial crosshair arm length, in data coordinates. Only used when
+        ``style`` is ``MarkerStyle.CROSSHAIR_LENGTH``.
```
ZLLentz, line 45: Comment: good catch documenting this- it was missing before :+1:

pcdswidgets/imaging/common/cam_marker.py
```diff
@@ -37,34 +57,55 @@ def __init__(
...
-        self._rebuild()
 
-    def detach(self) -> None:
-        """Remove all graphic items from the ViewBox."""
-        self._remove_items()
-        self._view_box = None
+        # One entry per attached ViewBox: {"view_box", "offset_x", "offset_y", "items"}
+        self._attachments: list[dict] = []
+
+    def attach(self, view_box, offset: tuple[float, float] = (0.0, 0.0)) -> None:
```
ZLLentz: Nitpick: is there a `pyqtgraph` class you can pick as the type annotation for `view_box` here and elsewhere?

pcdswidgets/imaging/common/centroid_tracker_full.py
```diff
@@ -0,0 +1,689 @@
...
+_FWHM_TO_SIGMA = 2.355  # FWHM ≈ 2.355 x sigma for a Gaussian
+_MIN_ROI_SIZE = 10  # pixels
+
+# A brand-new PyDMChannel's first connect can silently never call
+# connection_slot, so retry a few times before giving up.
+_CONNECT_RETRY_INTERVAL_MS = 2000
+_CONNECT_RETRY_MAX_ATTEMPTS = 5
+
+
+class _PVWriter(QObject):
```
ZLLentz, line 53: Comment: this is a lot of boilerplate for what amounts to a single caput handler, but it's probably justified because of the connection checking and the writes via qt signal, which won't block the main thread. My first instinct was to suggest simplifying this but on further consideration I think it could stay.

The main alternative is to create `PyDMChannel` objects directly in `CentroidTrackerFull`.

ZLLentz (commented):
No blockers from me- mostly looks good.

Please consider and reply to anything that isn't a nitpick and I'm happy to approve after some small discussion.

I'm also going to add @ipotanin as a reviewer since he made the previous versions of most of these files. I'm curious for others' thoughts as well.

pcdswidgets/imaging/common/cam_marker.py
moyinoluwa-10: It should be okay as is. I just added new styles. I don't think there's any key difference between a "line" and "shape". New shapes can just be assigned to a number. It all falls under the "marker style".

pcdswidgets/imaging/common/cam_marker.py
moyinoluwa-10: Marker styles screenshots have been added to the PR description.

pcdswidgets/imaging/common/cam_marker.py
moyinoluwa-10: Yes, there is. I've added it and it will be pushed with any other updates required before merging.

ZLLentz (dismissed):
I'll leave my approval, but I suggest getting at least one pair of eyes on this from someone who will use it themselves!

pcdswidgets/imaging/common/centroid_tracker_full.py
```diff
@@ -0,0 +1,689 @@
...
+
+    def _connect_value_labels(self):
+        """Hook the centroid/sigma/threshold PyDMLabels to track their live values (wired up once)."""
+        if self._labels_connected:
+            return
+
+        self._wrap_value_changed(self.centroid_x_label, lambda value: self._on_centroid_changed(value, "x"))
+        self._wrap_value_changed(self.centroid_y_label, lambda value: self._on_centroid_changed(value, "y"))
+        self._wrap_value_changed(self.sigma_x_label, lambda value: self._on_sigma_changed(value, "x"))
+        self._wrap_value_changed(self.sigma_y_label, lambda value: self._on_sigma_changed(value, "y"))
```
aberges-SLAC, line 256: Any particular advantage over using `lambda` instead of `partial`?

pcdswidgets/imaging/common/cam_marker.py
```diff
@@ -123,65 +178,114 @@ def width(self) -> int:
...
+        if not items:
             return
 
+        x = self.x - attachment["offset_x"]
+        y = self.y - attachment["offset_y"]
+
         if self._style == MarkerStyle.INFINITE_LINES:
-            self._items[0].setValue(self.y)  # horizontal
-            self._items[1].setValue(self.x)  # vertical
+            items[0].setValue(y)  # horizontal
```
aberges-SLAC, line 255: y is horizontal and x is vertical universally in this widget? I'm a little perplexed as to why it rotates, but the widget _does_ work so I guess that is true, I'm just super curious why?

pcdswidgets/imaging/common/colormap_intesity_control_full.py
```diff
@@ -65,11 +69,17 @@ def link_parent_widgets(self, parent: any) -> None:
 
         Called by the parent widget at adoption time. Links the histogram
         to the image item and syncs the UI state to current image settings.
+
```
ipotanin, line 72: Just thinking out loud, I'm fine with leaving as-is and no action items here.

I could see a scenario where the histogram on the ROI view would be more useful. In my test example the full screen image is mostly dark, so my full view histogram is a little useless.

pcdswidgets/imaging/common/centroid_tracker_full.py
```diff
@@ -0,0 +1,689 @@
...
+        PyDMLabel has no public "new value" signal, so value_changed is
+        wrapped in place.
+        """
+        original = label.value_changed
+
+        def wrapped(new_value, _original=original, _callback=callback):
+            _original(new_value)
+            _callback(new_value)
+
+        label.value_changed = wrapped
```
aberges-SLAC, line 274: This is pretty clever, I may steal it for use in other widgets

pcdswidgets/imaging/common/centroid_tracker_full.py
moyinoluwa-10: None to my knowledge.

pcdswidgets/imaging/common/centroid_tracker_full.py
```diff
@@ -0,0 +1,689 @@
...
+        self._sync_threshold_for_live_modes()
+
+    def _on_threshold_rbv_changed(self, value: float) -> None:
+        """Track the live CentroidThreshold_RBV, used to seed the raw-mode line edit."""
+        if value is None:
+            return
+        try:
+            self._threshold_rbv = float(value)
+        except (ValueError, TypeError):
+            logger.warning(f"Invalid CentroidThreshold_RBV received: {value}")
```
aberges-SLAC, line 358: I see this idiom pop up a _lot_ in this widget and I wonder if we could've captured it in a function that accommodates casting the value when necessary:
```
def try_to_set_value(self, value: any, attr_name: str, cast: int | float | str = None):
    if value is None:
        return
    try:
        if not cast:
           setattr(self, attr_name, value)
        else:
           setattr(self, attr_name, cast(value))
...
```

This wouldn't work as an exact drop in if you're also checking for "x" or "y" here and maybe this is crazy on my end to suggest

pcdswidgets/imaging/common/centroid_tracker_full.py
```diff
@@ -0,0 +1,689 @@
...
+        Cosmetic only (never writes to EPICS), so it's safe to call during init.
+        """
+        self.threshold_value_edit.setEnabled(mode != _THRESHOLD_MODE_ONE_OVER_E2)
+        if mode == _THRESHOLD_MODE_PERCENT:
+            self._threshold_value_validator.setRange(0.0, 100.0, 4)
+            self.threshold_value_edit.setPlaceholderText("% of max")
+            if not self.threshold_value_edit.text():
+                self.threshold_value_edit.setText("10")
+        elif mode == _THRESHOLD_MODE_RAW:
+            self._threshold_value_validator.setRange(0.0, 1e9, 4)
```
aberges-SLAC, line 393: Are we ever going to be more than max `uInt16` for a single pixel intensity?

pcdswidgets/imaging/common/centroid_tracker_full.py
moyinoluwa-10: It's a good alternative to subclassing PyDMLabel

pcdswidgets/imaging/common/centroid_tracker_full.py
ZLLentz: I was curious about this in a broad sense. Based on this ruff rule: https://docs.astral.sh/ruff/rules/lambda-assignment/, I think the only "standard rule" is that it's improper to directly assign lambdas to variable names. That is, if you want to do something like:
```
x = lambda g: g**2
```
You should instead use a partial or define a full function.

If you want to directly pass it into a function call, it looks like it's acceptable to use any of a lambda, a partial, or a full function definition.

pcdswidgets/imaging/common/cam_marker.py
moyinoluwa-10: It's not that y is horizontal and x is vertical. It's that a horizontal line is positioned by its y-coordinate, and a vertical line is positioned by its x-coordinate.

pcdswidgets/imaging/common/centroid_tracker_full.py
moyinoluwa-10: No we are not. This validator exists to validate our input based on the `threshold_mode_combo` value (raw vs percentage).

pcdswidgets/imaging/common/centroid_tracker_full.py
moyinoluwa-10: I'm not sure if there's a clean way to implement this, but I'll see if optimization is possible.

pcdswidgets/imaging/common/collapsible_section.py
```diff
@@ -28,6 +28,8 @@ class CollapsibleSection(QWidget):
 
     """
 
+    toggled = Signal(bool)  # emits the new expanded state (True = expanded)
```
ipotanin, line 31: Was this a specific request to track which tabs are folded (and then saved)? I'm guessing so. It seems like a little overkill to me

pcdswidgets/imaging/common/centroid_tracker_full.py
ipotanin: I actually have not heard of partials before. Learned something today.

pcdswidgets/ui/imaging/common/centroid_tracker_full.ui
```diff
@@ -0,0 +1,600 @@
...
+         <height>16777215</height>
+        </size>
+       </property>
+       <property name="text">
+        <string>Threshold</string>
+       </property>
+      </widget>
+     </item>
+     <item>
+      <widget class="QComboBox" name="threshold_mode_combo">
```
ipotanin, line 487: nitpick:
on first use of this widget I was a little confused that the readback is always in units of raw-intensity and even in percent mode.

I can also see auto draw ROI row being a little confusing
maybe change the button name  to "Auto Set ROI2" and move it before the No. of Widths control.

Feel free to ignore if you disagree, or you got no feedback from users
Image

ipotanin (dismissed):
LGTM, A couple of comments added but no blockers.

pcdswidgets/imaging/common/colormap_intesity_control_full.py
aberges-SLAC: I think this will be universally true for all far field cameras. Focal spots are really tight and we are wasting a lot of pixels

pcdswidgets/ui/imaging/common/centroid_tracker_full.ui
aberges-SLAC: I'd even accept `Recalc. Crop ROI` as a button name.

As for intensity, I suppose you could force `egu=%` into the `QLineEdit` by way of text update fstrings, but I don't feel strongly enough about this to push for it. 

On the contrary, having it be the same `a.u.` of intensity is more intuitive to me because I'm writing the same threshold value

pcdswidgets/imaging/common/colormap_intesity_control_full.py
moyinoluwa-10: To fix this issue is fairly simple. All you have to do is define the ROI view and the primary `image_view`. This makes the histogram depend on that for the source of truth and the same histogram is applied in the Full view.

pcdswidgets/imaging/common/collapsible_section.py
moyinoluwa-10: Yes, it is. It serves as part of the saved setting features I added so that a widget/app appears the same way it was left when reopened after closing.

pcdswidgets/imaging/common/centroid_tracker_full.py
aberges-SLAC: I'm pretty sure Python supports unbound methods as a variable, which means you can store your cast as an argument. Perhaps it's more clear if we use something like `partial` to do so? But for the basic types I thought this was supported.

This is definitely not a blocking issue but I've seen the idiom enough in this script that I thought it was worth mentioning

pcdswidgets/imaging/common/centroid_tracker_full.py
aberges-SLAC: As I mentioned in person but should probably take the time to document it here too:
The `intensity` passed to `Stats2` probably maxes out at `uint16`, but most of our cameras are 12 bit or 14 (?) bit mono. However, I haven't given a whole lot of thought to how Area Detector deals with binning on these 16-bit numbers.

Naively, I'd assume a bin of 4x4 max intensity pixels still clips to `655335`, but if the under-the-hood calc records are using floats, we might have 32-64 bits at our disposable. So we could conceivably hit `1e9` or even `2^32-1`. Regardless, I think this is a perfectly acceptable magic number

aberges-SLAC (approved):
I think if Zach wants to give it another go is totally welcome, but your latest commit was pretty harmless and just updating the type-hinting. I don't see any reason why it should invalidate prior approval.

Here's my ✅

ZLLentz (approved):
Nothing more from me- if Adam is happy and there weren't any other critical change requests, should be good to go. If there is more work to do we can make more PRs.

PR #136, opened 2026-08-07
first pass at setting macro properties after macros are available

KaushikMalapati:
## Description

In the pyuic generated form files, there are two functions - setupUi and retranslateUi, which both set properties of widgets that can include macros; setupUi calls retranslateUi (at least the first time) and is called during DesignerWidget.__init__. However, macros in this library are set after initialization but before the event loop starts meaning that widgets will need macros before we make them available. 

To remedy this, I have made a simple change - any line in setupUi or retranslateUi that would use a macro now places that line in a nested macro_setup or macro_retranslate function, called by a single shot qtimer with 0 delay - this means it is called as soon as the event loop starts, which is after macros are available. Finally, we call _updates_for_macro on all the macros for the form, as anything that was set previously using set_macro would have found nothing to expand.

## Motivation and Context

https://jira.slac.stanford.edu/browse/ECS-10624

## How Has This Been Tested?

Interactively. I saw the errors without my changes, and they went away after unstashing my changes and rebuilding.

Not sure if this needs any documentation

## Pre-merge Checklist
- [x] Description and screenshot and of these widgets working are added to the ECS Widget Catalog https://confluence.slac.stanford.edu/spaces/PCDS/pages/716925985/ECS+UI+UX+pcdswidgets+Widget+Catalog


KaushikMalapati:
@ZLLentz can you take a look at this draft? I'm a bit iffy on how I went about things, and I think there are some further optimizations I could make (parsing macro_setup and macro_retranslate for which macros I should call _updates_for_macro  on instead of doing all of them, making a function for each widget that the first _update_widget_for_macros that has all macros ready are defined calls) but I wanted to get a look before I go any further. Including the generated files so you can see how the the new Ui_Forms  would look

pcdswidgets/generated/common/tools/feature_finder_form.py
```diff
@@ -515,30 +518,45 @@ def retranslateUi(self, Form):
...
+            self.velocity_get.setChannel(_translate("Form", "ca://${motor}.VELO"))
+            self.detector_get.setChannel(_translate("Form", "ca://${detector}"))
+            self.position_get.setChannel(_translate("Form", "ca://${motor}.RBV"))
+            self.velocity_set.setChannel(_translate("Form", "ca://${motor}.VELO"))
+            self.position_set.setChannel(_translate("Form", "ca://${motor}.VAL"))
+            self.motor_status_led.setProperty("motor", _translate("Form", "${motor}"))
+            for macro in Form._macro_to_widget:
+                Form._updates_for_macro(macro)
+
+        QTimer.singleShot(0, macro_retranslate)
```
ZLLentz: Reviewing in the built file because it's easier to reason based on the generated code,

I'm not sure why this works- we're still setting all the channels to macro-laden values. I guess the immediate swap to fill the macro values saves us before PyDM/pyepics gets too far?

What happens if we skip this step entirely, e.g. remove these sets without calling `macro_retranslate`? My intuition is that the widgets all still work the same, getting their values from the macro setters in `DesignerWidget`. I think this must be happening immediately prior to `macro_retranslate` getting called in this iteration of the code, else `_updates_for_macro` wouldn't be helping here.

So I think the logic flow in this version is:

- setupUi/retranslateUi
- timer starts
- "motor" macro is set
- all channels get updated to the correct values
- timer fires
- all channels gets updated to the pre-macro-fill values
- all channels get put back to the correct values

Is it as simple as just removing the property sets and letting the "base" class handle filling and setting macro-ized properties?

KaushikMalapati: Tried commenting out `QTimer.singleShot(0, macro_retranslate)` so macro_retranslate isn't called, and the channels connect (as long as you set the protocol macro) but I don't understand why. If for example, this line
`self.position_set.setChannel(_translate("Form", "ca://${motor}.VAL"))`
is never run, when set_motor is called, wouldn't `self.position_set.channel` be None since it was never set, and therefore there is no macro value to expand for the channel? I feel like I'm missing something

KaushikMalapati: Looks like it's this part of _update_widget_for_macros
https://github.com/pcdshub/pcdswidgets/blob/699108c21cb21f912cb7c13af1eeb734d375c52f/pcdswidgets/builder/designer_widget.py#L197-L204
Didn't look too closely at _widget_to_pre_template before, but if we store what the pre-expansion property is going to be and expand that before setting the property for the widget, there's no reason to do it in setupUi or retranslateUi right?

KaushikMalapati:
I think this is ready for review. Adjusted some existing test cases and rebuilt everything

pcdswidgets/generated/common/tools/feature_finder_form.py
ZLLentz: I think so

pcdswidgets/builder/build.py
```diff
@@ -31,7 +31,10 @@ def build_uic(designer_ui: str, output_dir: str = ""):
             import_lines.append(line.replace("PyQt5", "qtpy"))
             continue
         if import_lines:
-            impl_lines.append(line)
+            if "${" not in line:
+                impl_lines.append(line)
+            elif impl_lines[-1].isspace():
+                del impl_lines[-1]
```
ZLLentz, line 37: Well isn't that a compact diff

pcdswidgets/tests/builder/test_designer_widget.py
```diff
@@ -94,15 +94,15 @@ def test_has_expected_macro_values(basic_test_widget: BuilderBasicTestWidget):
 
 
 def test_macro_substitution_labels(basic_test_widget: BuilderBasicTestWidget):
-    assert basic_test_widget.name_label.text() == "Name: ${NAME}"
-    assert basic_test_widget.num_label.text() == "Num: ${NUM}"
-    assert basic_test_widget.name_num_label.text() == "${NAME}:${NUM}"
+    assert basic_test_widget.name_label.text() == ""
```
ZLLentz, line 97: Forgot about the tests, thanks for updating them

ZLLentz (approved):
I don't think there's a simpler way to do it, I like this

ZLLentz:
Small aside: the reason why `retranslateUi` exists separately from `setupUi` is because qt has some built-in utilities for helping you make your ui work in multiple languages. I thought that was interesting.

pcdswidgets/tests/builder/test_designer_widget.py
KaushikMalapati: I also forgot until the checks failed

